PR #5, opened 2022-01-18
Rahwa-Ghebremichael/ JS-1/Week-4

rahwafesseha:


Monika39-maker (commented):
your codes are almost similar to mine.

1-exercises/F-array-forEach/exercise.js
```diff
@@ -7,7 +7,19 @@
...
 let arr = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
+  arr.forEach((element) => {
+    console.log(
+      element % 3 === 0 && element % 5 === 0
+        ? "FizzBuzz"
+        : element % 3 === 0
+        ? "Fizz"
+        : element % 5 === 0
+        ? "Buzz"
+        : element
```
mahsa2, line 20: It's nice that you're using the conditional (ternary) operator here for if/else. In most companies, they set a rule not to use `conditional (ternary) operator` if it is going to be nested. The recommendation is to use it if it is a simple one, like below. Usually, bugs show up in the nested operators and it's hard to maintain the code:
```isMember ? '$2.00' : '$10.00'```

File: 1-exercises/E-array-map/exercise.js [hunk 2]
```diff
@@ -3,7 +3,10 @@
 
 let numbers = [0.1, 0.2, 0.3, 0.4, 0.5];
 
-let numbersMultipliedByOneHundred; // complete this statement
+function multiplyBy100(num) {
+  return num * 100;
+}
+let numbersMultipliedByOneHundred = numbers.map(multiplyBy100); // complete this statement
```
mahsa2, line 9: What you have here is correct. Just to introduce other formats to you:
```
numbers.map(num => num * 100)
```
```
numbers.map(function (num) {
  return num * 100;
})
```

2-mandatory/1-create-functions.js
```diff
@@ -44,7 +54,15 @@ Write a function that:
...
 */
 
-function formatPercentage() {
+function formatPercentage(numbers) {
+  return numbers.map((number) => {
+    if (number > 100){
+      return "100%";
+    }
+    else { 
+       return Number(number.toFixed(2)) + "%";
```
mahsa2, line 63: Nice, also you can use template literals:
``${number.toFixed(2)}%``

2-mandatory/6-journey-planner.js
```diff
@@ -81,7 +83,13 @@ function getTransportModes() {}
 
   Hint: Use the corresponding array method to decide if an element is included in an array.
 */
-function isAccessibleByTransportMode() {}
+function isAccessibleByTransportMode(ArrayOfTransportMode, stringOfTransportMode) {
+  if(ArrayOfTransportMode.includes(stringOfTransportMode)) {
+    return true;
+  }else{
+    return false;
+  }
```
mahsa2, line 91: Well done. Here because `ArrayOfTransportMode.includes(stringOfTransportMode)` returns a `bool` we ca n dirrectly return its results and avoid returning `true/false` or using the `if/else` statements.
```
return ArrayOfTransportMode.includes(stringOfTransportMode);
```

mahsa2 (commented):
Pretty neat 💯 You've learned a lot by now, that's amazing 🥳

PR #6, opened 2022-01-18
London Class 8 - Craig D'Silva - JavaScript Core 1 - Week 4

craig-dsilva:
I couldn't find a solution for the final mandatory check.

### DIFF
--- a/1-exercises/C-array-every/exercise.js
+++ b/1-exercises/C-array-every/exercise.js
@@ -6,11 +6,7 @@ let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
 function isStudents(student) {
-  for (element of students) {
-    if (element === student) {
-      return true;
-    }
-  }
+  return students.includes(student);
 }
 
 let groupIsOnlyStudents = group.every(isStudents); // complete this statement
--- a/1-exercises/C-array-every/exercise.js
+++ b/1-exercises/C-array-every/exercise.js
@@ -5,7 +5,11 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+function isStudents(student) {
+  return student === students;
+}
+
+let groupIsOnlyStudents = group.every(isStudents); // complete this statement
 
 if (groupIsOnlyStudents) {
   console.log("The group contains only students");
--- a/1-exercises/C-array-every/exercise.js
+++ b/1-exercises/C-array-every/exercise.js
@@ -6,7 +6,11 @@ let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
 function isStudents(student) {
-  return student === students;
+  for (element of students) {
+    if (element === student) {
+      return true;
+    }
+  }
 }
 
 let groupIsOnlyStudents = group.every(isStudents); // complete this statement
--- a/1-exercises/D-array-filter/exercise.js
+++ b/1-exercises/D-array-filter/exercise.js
@@ -9,7 +9,7 @@
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
 let pairsByIndex = pairsByIndexRaw.filter(
-  (value) => value instanceof Array && value.length > 1
+  (value) => value instanceof Array && value.length === 2
 ); // Complete this statement
 
 let students = ["Islam", "Lesley", "Harun", "Rukmini"];
--- a/1-exercises/D-array-filter/exercise.js
+++ b/1-exercises/D-array-filter/exercise.js
@@ -8,7 +8,9 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+let pairsByIndex = pairsByIndexRaw.filter(
+  (value) => value instanceof Array && value.length > 1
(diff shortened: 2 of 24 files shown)


1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,11 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+function isStudents(student) {
+  return student === students;
+}
```
Selchuk-Karakus: This function isn't exactly doing what we expect it to do. If you remove the item "Daniel" from the group array and run the function, unfortunately, it breaks.  You just need to replace the === operator with an Array method that checks to see if a value is **included** among the array it's doing the check on rather than checking whether they are equal.  You're very close.

Selchuk-Karakus (changes requested):
Keep going Craig, you're very close!

1-exercises/C-array-every/exercise.js
craig-dsilva: I fixed it. Thanks for letting me know.

1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,15 @@
...
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+function isStudents(student) {
+  for (element of students) {
+    if (element === student) {
+      return true;
+    }
+  }
+}
```
DomVinyard: We can also do this without an explicit loop

```
function isStudents(student) {
  return students.includes(student)
 }
```

1-exercises/D-array-filter/exercise.js
```diff
@@ -8,7 +8,9 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+let pairsByIndex = pairsByIndexRaw.filter(
+  (value) => value instanceof Array && value.length > 1
```
DomVinyard: I think `value.length === 2` is better than `value.length > 1`. Your solution would fail if a test contained an array with 3 items.

1-exercises/F-array-forEach/exercise.js
```diff
@@ -9,6 +9,21 @@
...
+    return "FizzBuzz";
+  }
+  if (value % 3 === 0) {
+    return "Fizz";
+  }
+  if (value % 5 === 0) {
+    return "Buzz";
+  }
+  return value;
+});
```
DomVinyard, line 23: FizzBuzz is great! Nice solution

```
arr.map(x=>(x%3?"":"Fizz")+(x%5?"":"Buzz")||x)
```

 😅

2-mandatory/1-create-functions.js
```diff
@@ -24,7 +33,11 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
+function tidyUpString(arr) {
+  let tidyArr = arr.map((element) =>
+    element.toLowerCase().replace("/", "").trim()
+  );
+  return tidyArr;
```
DomVinyard, line 40: 👍

2-mandatory/1-create-functions.js
```diff
@@ -44,7 +60,21 @@ Write a function that:
...
+
+function numChecker(num) {
+  if (num > 100) {
+    return "100%";
+  } else if (Number.isInteger(num)) {
+    return num + "%";
+  } else if (num < 1) {
+    return num.toFixed(2) + "%";
+  } else {
+    return num.toFixed(1) + "%";
```
DomVinyard, line 76: I remember, this was to satisfy the tests right. Reminder for me to chase this up with the syllabus team

2-mandatory/5-eligible-students.js
```diff
@@ -7,7 +7,14 @@
...
 
-function getEligibleStudents() {}
+function getEligibleStudents(attendanceArr) {
+  let filteredArr = attendanceArr.filter((element) => element[1] >= 8);
+  let loopedArr = [];
+  for (let element of filteredArr) {
+    loopedArr.push(element[0]);
+  }
+  return loopedArr;
+}
```
DomVinyard: I think I'd favour an array function here rather than a loop

```suggestion
function getEligibleStudents(attendanceArr) {
  let filteredArr = attendanceArr.filter((element) => element[1] >= 8);
  return filteredArr.map(element => element[0])
}
```

2-mandatory/6-journey-planner.js
```diff
@@ -81,7 +85,13 @@ function getTransportModes() {}
...
   Hint: Use the corresponding array method to decide if an element is included in an array.
 */
-function isAccessibleByTransportMode() {}
+function isAccessibleByTransportMode(transportArr, transport) {
+  if (transportArr.includes(transport)) {
+    return true;
+  } else {
+    return false;
+  }
+}
```
DomVinyard: for these we can simplify it by just returning the boolean straight from the result of the `includes` function

```suggestion
function isAccessibleByTransportMode(transportArr, transport) {
  return transportArr.includes(transport)
}
```

3-extra/2-sorting-algorithm.js
```diff
@@ -14,7 +14,25 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
...
+    }
+  }
+  let splittedArr = newArr;
+  for (let i = 0; i < splittedArr.length; i++) {
+    for (let j = 0; j < splittedArr.length; j++) {
+      if (splittedArr[j] > splittedArr[i]) {
+        check = splittedArr[i];
+        splittedArr[i] = splittedArr[j];
+        splittedArr[j] = check;
+      }
```
DomVinyard, line 31: I'm not sure I follow the logic here, can you talk me through it on our call

PR #8, opened 2022-01-19
LONDON CLASS 8 - Daniel Piga - JS CORE1 WEEK 4

pigos4:
@kolomiets

### DIFF
--- a/1-exercises/B-array-some/exercise.js
+++ b/1-exercises/B-array-some/exercise.js
@@ -7,7 +7,7 @@
 */
 
 let pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
-
+if (pairsByIndex.some((value) => value === null)) process.exit(1);
 // If there is a null value in the array exit the program with the error code
 // https://nodejs.org/api/process.html#process_process_exit_code
 // process.exit(1);
--- a/1-exercises/C-array-every/exercise.js
+++ b/1-exercises/C-array-every/exercise.js
@@ -5,7 +5,9 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents = group.every((singlestudent) =>
+  students.includes(singlestudent)
+); // complete this statement
 
 if (groupIsOnlyStudents) {
   console.log("The group contains only students");
--- a/1-exercises/D-array-filter/exercise.js
+++ b/1-exercises/D-array-filter/exercise.js
@@ -8,7 +8,9 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+let pairsByIndex = pairsByIndexRaw
+  .filter((item) => item)
+  .filter((item) => item.length === 2); // Complete this statement
 
 let students = ["Islam", "Lesley", "Harun", "Rukmini"];
 let mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
--- a/1-exercises/F-array-forEach/exercise.js
+++ b/1-exercises/F-array-forEach/exercise.js
@@ -8,7 +8,17 @@
 */
 
 let arr = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
-
+arr.forEach((num) => {
+  if (num % 3 === 0 && num % 5 === 0) {
+    console.log("FizzBuzz");
+  } else if (num % 3 === 0) {
+    console.log("Fizz");
+  } else if (num % 5 === 0) {
+    console.log("Buzz");
+  } else {
+    console.log(num);
+  }
+});
 /* EXPECTED OUTPUT */
 
 /*
--- a/1-exercises/G-array-methods/exercise.js
+++ b/1-exercises/G-array-methods/exercise.js
@@ -4,7 +4,7 @@
 */
 
 let numbers = [3, 2, 1];
-let sortedNumbers; // complete this statement
+let sortedNumbers = numbers.sort((a, b) => a - b); // complete this statement
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/1-exercises/G-array-methods/exercise2.js
+++ b/1-exercises/G-array-methods/exercise2.js
@@ -7,7 +7,7 @@
 let mentors = ["Daniel", "Irina", "Rares"];
 let students = ["Rukmini", "Abdul", "Austine", "Swathi"];
 
-let everyone; // complete this statement
+let everyone = [...mentors, ...students]; // complete this statement
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/1-exercises/H-array-methods-2/exercise.js
+++ b/1-exercises/H-array-methods-2/exercise.js
@@ -15,8 +15,9 @@ let everyone = [
   "Swathi",
 ];
 
-let firstFive; // complete this statement
-let lastFive; // complete this statement
+let firstFive = everyone.filter((v, index) => index <= 4); // complete this statement
+
+let lastFive = everyone.filter((v, index, array) => index >= array.length - 5); // complete this statement
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/1-exercises/H-array-methods-2/exercise2.js
+++ b/1-exercises/H-array-methods-2/exercise2.js
@@ -7,7 +7,13 @@
   Tip: use the string method .split() and the array method .join()
 */
 
-function capitalise(str) {}
+function capitalise(str) {
+  let strToArray = [...str];
+  return strToArray
+  .shift()
+  .toUpperCase()
+  .concat(strToArray.join(""));
+}
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/1-exercises/I-string-replace/exercise.js
+++ b/1-exercises/I-string-replace/exercise.js
@@ -13,7 +13,11 @@
 let story =
   "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
 
-let result = story.replace("", "");
+let result = story
+  .replace(/dogs/g, "cats")
+  .replace(10, 100000)
+  .replace(/day/g, "night")
+  .replace("great", "brilliant");
 
 /* EXPECTED OUTPUT */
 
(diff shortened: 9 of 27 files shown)


1-exercises/B-array-some/exercise.js
```diff
@@ -7,7 +7,7 @@
 */
 
 let pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
-
+if (pairsByIndex.some((value) => value === null)) process.exit(1);
```
kolomiets, line 10: Use one-line 'if' statements carefully. It is the only exception in [Google's JavaScript Style Guide](https://google.github.io/styleguide/jsguide.html#formatting-braces-all) but I know that many programmers still frown upon it :)

1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,9 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents = group.every((singlestudent) =>
```
kolomiets, line 8: Maybe my personal style choice (your may be different and I like the current version!)

I tend to use short parameter names for the lambdas (and I never do this for "proper" named functions):

```javascript
let groupIsOnlyStudents = group.every(e => students.includes(e));
```

The reasoning is that lambdas need to be very simple and short and usually they can be easily understood even with short parameter names. Named functions, in contrast, are typically larger and a good name for parameter really helps.

1-exercises/D-array-filter/exercise.js
```diff
@@ -8,7 +8,9 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+let pairsByIndex = pairsByIndexRaw
```
kolomiets, line 11: Take a look at [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

A shorter (and maybe more readable) version would be:

```javascript
let pairsByIndex = pairsByIndexRaw.filter(p => p?.length === 2)
```

1-exercises/F-array-forEach/exercise.js
```diff
@@ -8,7 +8,17 @@
 */
 
 let arr = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
-
+arr.forEach((num) => {
```
kolomiets, line 11: There is a common wisdom (which, as a good programmer, you can go against when you need) - a function should have a single exit point (aka return statement). Having multiple return statements in a function generally makes it more difficult to understand and reason about. Consider this version:

```javascript
arr.forEach((number) => {
  let message = number
  if (number % 15 === 0) {
    message = "FizzBuzz";
  } else if (number % 3 === 0) {
    message = "Fizz";
  } else if (number % 5 === 0) {
    message = "Buzz";
  }
  console.log(message);
});
```

Doing it this way, we also reduce the code repetition (console.log calls) - so, for example, if we want to change the way we output the message, we need to change only one line, not multiple lines as before.

1-exercises/G-array-methods/exercise.js
```diff
@@ -4,7 +4,7 @@
 */
 
 let numbers = [3, 2, 1];
-let sortedNumbers; // complete this statement
+let sortedNumbers = numbers.sort((a, b) => a - b); // complete this statement
```
kolomiets, line 7: You can use `sort()` - it is ascending sort by default. The current version may be confusing - a reader needs to stop for a second to understand it.

1-exercises/G-array-methods/exercise2.js
```diff
@@ -7,7 +7,7 @@
 let mentors = ["Daniel", "Irina", "Rares"];
 let students = ["Rukmini", "Abdul", "Austine", "Swathi"];
 
-let everyone; // complete this statement
+let everyone = [...mentors, ...students]; // complete this statement
```
kolomiets, line 10: What's better - your solution or `mentors.concat(students)`? Why?

1-exercises/H-array-methods-2/exercise.js
```diff
@@ -15,8 +15,9 @@ let everyone = [
   "Swathi",
 ];
 
-let firstFive; // complete this statement
-let lastFive; // complete this statement
+let firstFive = everyone.filter((v, index) => index <= 4); // complete this statement
```
kolomiets, line 18: Clever :) It is a very common operation and there is a method for it: [slice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice)

Solution would be considerable easier to read.

1-exercises/H-array-methods-2/exercise.js
```diff
@@ -15,8 +15,9 @@ let everyone = [
...
+
+let lastFive = everyone.filter((v, index, array) => index >= array.length - 5); // complete this statement
```
kolomiets, line 20: Not very efficient as you need to iterate whole array to get the last 5 elements. Consider how slow it's going to be if you have 1 million elements. 

Slice is generally preferred: 

```javascript
everyone.slice(-5)
```

1-exercises/H-array-methods-2/exercise2.js
```diff
@@ -7,7 +7,13 @@
   Tip: use the string method .split() and the array method .join()
 */
 
-function capitalise(str) {}
+function capitalise(str) {
+  let strToArray = [...str];
+  return strToArray
+  .shift()
```
kolomiets, line 13: It works but I'd argue that it's hard to read and understand as you replace a sequence of elements with a single element when you "chain" function invocations. Chaining works and looks great when you operate on sequences (using filter, map, reduce, etc). 

Consider this version:

```javascript
function capitalize(str) {
    let letters = str.split("")
    letters[0] = letters[0].toUpperCase()
    return letters.join("")
}
```

1-exercises/I-string-replace/exercise.js
```diff
@@ -13,7 +13,11 @@
 let story =
   "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
 
-let result = story.replace("", "");
+let result = story
```
kolomiets, line 16: 👍 Call chaining is a very powerful technique - it's may be difficult to grasp, but once you get it, you see it everywhere :)

2-mandatory/1-create-functions.js
```diff
@@ -24,7 +26,8 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
+function tidyUpString(array) {
+  return array.map((item) => item.replace(/^\s+|\/|\s+$/g, "").toLowerCase());
```
kolomiets, line 30: Be **very** careful with regexes. The syntax is obscure and even the author of a complex regex stops understanding after 5 minutes :) My advice would be to limit use of regexes if you can, especially for simple cases.

Consider use of [trim](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim):

```javascript
item.trim().replace("/", "").toLowerCase()
```

I'm almost certain that non-regex version is much faster as well (try to prove/disprove this statement - measure the performance of these two alternative solutions).

2-mandatory/3-bush-berries.js
```diff
@@ -22,7 +22,10 @@
 */
 
 function isBushSafe(berryArray) {
```
kolomiets, line 24: Maybe, just maybe - there is a right or wrong answer here - this may be more readable and easier to support going forward:

```javascript
  const safe = berryArray.every((berry) => berry === "pink")
  return safe 
    ? "Bush is safe to eat from"
    : "Toxic! Leave bush alone!";
}
```

2-mandatory/4-space-colonies.js
```diff
@@ -15,7 +15,9 @@
   
 */
 
-function getSettlers() {}
+function getSettlers(array) {
+  return array.filter((value) => value.match(/^A/g) && value.match(/family$/g));
```
kolomiets, line 19: Hm... I'd argue that non-regex version is far easier to read and works faster:

```javascript
 value[0] === "A" && value.includes("family")
```

Again, regex is a great powerful tool - but often there is an easier route. Look for it! :)

2-mandatory/6-journey-planner.js
```diff
@@ -18,10 +18,9 @@
 */
 
 function checkCodeIsThere(stringText) {
-  let magicWord = "code";
-  //edit code below
-  if (stringText) {
-    return stringText;
+  let index = stringText.match(/code/);
```
kolomiets, line 21: I'd argue that ```stringText.indexOf(magicWord)``` may be a better fit for the job

2-mandatory/6-journey-planner.js
```diff
@@ -64,7 +63,11 @@ function checkCodeIsThere(stringText) {
 
   Hint: Use the corresponding array method to split the array.
 */
-function getTransportModes() {}
+function getTransportModes(array) {
```
kolomiets, line 66: Consider slice: 

```javascript
return array.slice(1);
```

2-mandatory/6-journey-planner.js
```diff
@@ -92,7 +97,10 @@ function isAccessibleByTransportMode() {}
    - Returns the name of the location
       e.g: "Tower Bridge"
 */
-function getLocationName() {}
+function getLocationName(locationAndTransports) {
+  let locationAndTransportsCopy = [...locationAndTransports];
```
kolomiets, line 101: Hm... so we create the whole new array to return the first element. Why don't we

```javascript
return locationAndTransportModes[0];
```

2-mandatory/6-journey-planner.js
```diff
@@ -122,7 +130,12 @@ function getLocationName() {}
   Advanced challange: try to use arrow function when invoking an array method.
 */
 function journeyPlanner(locations, transportMode) {
```
kolomiets, line 132: Make use of the functions you defined above:

```javascript
  return locations
    .filter(location => isAccessibleByTransportMode(location, transportMode))
    .map(location => getLocationName(location));
```

3-extra/1card-validator.js
```diff
@@ -0,0 +1,44 @@
...
+
+// oddLastNumber check if the last number is odd
+const oddLastNumber = (number) => /[02468]$/.test(number);
+
+// notOnlyOneNumber add all the number in a Set and if is more than one mean is not only one value
+const notOnlyOneNumber = (number) => new Set([...String(number)]).size !== 1;
+
+// moreThanSixteen check if sum of all the number is more than 16
+const moreThanSixteen = (number) =>
+  [...String(number)].reduce((a, b) => +a + +b) > 16;
```
kolomiets, line 14: `+a + +b` looks awful! :) Why don't you use a simpler version: `a + b`?

pigos4: I did like this because are strings, to convert the string to number.  
 Number(a)+Number(b)
maybe this is better?

PR #15, opened 2022-01-20
Ldn8 - Saman Zahedi - JS1 Week 4 

SamanZahedi:


2-mandatory/7-lane-names.js
```diff
@@ -6,7 +6,11 @@
   HINT: string and array methods that could be helpful (indexOf, filter)
 */
 
-function getLanes() {}
+function getLanes(streetNames) {
+  return streetNames.filter(
+    (streetName) => streetName.indexOf("Lane") === streetName.length - 4
```
beckyf0keefe, line 11: This works fine, but can you think of a way to do this using one of the methods that we learned this week?

beckyf0keefe (commented):
Overall looking good, couldn't see any issues with the logic you were using, only suggestion to make your code more understandable is to try and use some more descriptive names than "elem", some of the time its fine but in other places it could help the code be more understandable

PR #16, opened 2022-01-20
 Ezgi-Gunes-JavaScript-Core-1-Coursework-Week4-London8

Ezgi3406:


1-exercises/B-array-some/exercise.js
```diff
@@ -8,8 +8,16 @@
 
 let pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 
+let exit = pairsByIndex.some(k => k==null);
+
+if (exit){
+  return process.exit(1)
```
Selchuk-Karakus, line 14: Great use of process.exit method.  You have obviously read the documentation linked.

1-exercises/H-array-methods-2/exercise.js
```diff
@@ -15,8 +15,8 @@ let everyone = [
   "Swathi",
 ];
 
-let firstFive; // complete this statement
-let lastFive; // complete this statement
+let firstFive=everyone.slice(0,5) // complete this statement
+let lastFive=everyone.slice(2) // complete this statement
```
Selchuk-Karakus, line 19: This gives us the correct answer back in this use case, however, if the array length was greater, the logic here unfortunately breaks.  I would consider rewriting this so we are explicitly asking for the last 5 items regardless of however long the array may be.

1-exercises/J-string-substring/exercise2.js
```diff
@@ -14,11 +14,11 @@ let names = [
...
-names[0] = names[0].substring();
-names[1] = names[1].substring();
-names[2] = names[2].substring();
-names[3] = names[3].substring();
-names[4] = names[4].substring();
+names[0] = names[0].substring(0,6);
+names[1] = names[1].substring(0,7);
+names[2] = names[2].substring(0,4);
+names[3] = names[3].substring(0,4);
+names[4] = names[4].substring(0,5);
```
Selchuk-Karakus, line 21: This does return us the correct answer to pass the test. However, the arguments in the substring are hardcoded which means that if the test conditions change the logic, unfortunately, will break.  We could use indexOf() and split() here to make our logic robust, so, it's consistent with its output no matter what the test conditions are. By the way, I fell for the same mistake if it's any consolation : )

Selchuk-Karakus (commented):
Great job Ezgi!

1-exercises/A-array-find/exercise.js
```diff
@@ -16,8 +16,10 @@ let names = [
   "Karim",
   "Ahmed",
 ];
-
-let longNameThatStartsWithA = findLongNameThatStartsWithA(names);
+  function findLongNameThatStartsWithA(names){
+    return names.find(name => name.length > 7);
```
jstadnik, line 20: This checks for long names, but doesn't check that the names start with "A". For instance, if the input list contained "Gertrude", then that name would be included in the result.

1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,7 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents = group.every(e=>!students.filter(e=>e)) // complete this statement
```
jstadnik, line 8: I'm not sure this does what you intend. It's pretty much always going to return `false`. If you remove `"Daniel"` from `group` above, so that the group _is_ only students, the code still prints `"The group does not contain only students"`.

It's a bit complicated, so let's break it down:
- `students.filter(e => e)` will return all the elements of the `students` array that evaluate to `true` when considered as booleans (true/false values). The elements are strings, and JavaScript converts strings to booleans by checking if the string is empty and returning `false` if the string is empty, otherwise `true`. So all strings are `true` except for the empty string `""`. This means that, unless there's a student named `""`, this filter will just return the original array.
- `!arr`, where `arr` is an array, will similarly try to interpret the array as a boolean by checking if the array is empty. Assuming you have more than one student, `!students.filter(e=>e)` is always false, because the array will be non-empty.

Therefore, this is going to behave the same way as `group.every(e=>false)`! I think what you were looking for here was something like `group.every(e=> <something that evaluates to true if e is in students, otherwise false>)`. Thinking about the array methods you know, do you have an idea for what the "something" could be?

1-exercises/D-array-filter/exercise.js
```diff
@@ -8,7 +8,16 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+// let pairsByIndex= pairsByIndexRaw.filter(k => k!=null && k.length==2);
+ let pairsByIndex= pairsByIndexRaw.filter(k =>Array.isArray(k) && k.length==2);
```
jstadnik, line 12: Great use of `filter`!

1-exercises/H-array-methods-2/exercise2.js
```diff
@@ -7,7 +7,12 @@
   Tip: use the string method .split() and the array method .join()
 */
 
-function capitalise(str) {}
+function capitalise(str) {
+  return str
+  .split(" ")
+  .map(() => str.charAt(0).toUpperCase() + str.slice(1))
+  .join(" ");
+}
```
jstadnik, line 15: Good start! Does it work, though? Shouldn't the anonymous function in map take in a word as argument, rather than capitalise `str`?

1-exercises/I-string-replace/exercise.js
```diff
@@ -10,10 +10,16 @@
   "I like cats. One night I went to the park and I saw 100000 cats. It was a brilliant night."
 */
 
-let story =
-  "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
-
-let result = story.replace("", "");
+let story ="I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
+let regex =/\b(?:dogs|10|day|great)\b/gi
```
jstadnik, line 14: Clever use of regular expressions here! Way to go above and beyond.

1-exercises/J-string-substring/exercise3.js
```diff
@@ -7,9 +7,9 @@
 */
 
 let statement = "I do not like programming";
-
-let result = "";
-
+let start = statement.substring(0, statement.indexOf("not"));
+let end = statement.substring(statement.indexOf("not") + "not ".length, statement.length - 1);
+let result = start + end;
```
jstadnik, line 12: Nice use of `indexOf` here to make the logic flexible! Now even if you change the structure of the sentence above to e.g. "I really do not like programming", it'll still remove the "not" correctly :+1:

2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,18 @@ Write a function that:
...
 }
 
 /*
 Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(array) {
+  return [...array].sort();
```
jstadnik, line 17: Nice concise way to write this!

2-mandatory/1-create-functions.js
```diff
@@ -44,7 +54,20 @@ Write a function that:
...
+    .map((number) => {
+      // if/else statement
+      if (number > 100) {
+        return 100;
+      } else {
+        return number;
+      }
+
+      // ternary
+      // return number > 100 ? 100 : number;
```
jstadnik, line 68: You could also use `Math.min` for this kind of logic (e.g. `Math.min(number, 100)`).

2-mandatory/2-oxygen-levels.js
```diff
@@ -11,7 +11,21 @@
...
 */
 
-function findSafeOxygenLevel() {}
+ function findSafeOxygenLevel(array) {
+    return array.filter((item) => {
+      let num = item.replace("%", "");
+      if (item.includes("%") && num > 19.5 && num < 23.5) {
+        return item + "%";
+      }
+    })[0];
```
jstadnik, line 20: This works, but `array.filter(<func>)[0]` can be written more simply as `array.find(<func>)`.

2-mandatory/5-eligible-students.js
```diff
@@ -7,7 +7,9 @@
   - Returns an array containing only the names of the who have attended AT LEAST 8 classes
 */
 
-function getEligibleStudents() {}
+function getEligibleStudents(array) {
+  return array.filter((item) => item[1] >= 8).map((item) => item[0]);
```
jstadnik, line 11: Nice work with array methods here!

2-mandatory/6-journey-planner.js
```diff
@@ -16,17 +16,20 @@
...
   let magicWord = "code";
   //edit code below
   if (stringText) {
     return stringText;
+  }
+  if (stringText.includes(magicWord)) {
+    return stringText.indexOf(magicWord);
   } else {
     return "Not found";
   }
```
jstadnik, line 29: You've got the right idea here, but I think you need to remove the original placeholder code! Like I explained before, something like `if` that converts a string to a boolean will always interpret the string as `true` if it has any text at all; as a result, almost everything you pass in here will return the original string and never reach the code below.
```suggestion
  if (stringText.includes(magicWord)) {
    return stringText.indexOf(magicWord);
  } else {
    return "Not found";
  }
```

jstadnik (commented):
Great work! You seem to have a good handle on array methods :+1:

Something to think about might be to add extra test cases to the code you're given so that you can check the code more thoroughly -- for instance, in the "only students" example, it might be helpful to write another test that does have only students.

PR #19, opened 2022-01-20
London-Class 8/Laleh Koupaei/Java Script-Core1-Week4

lalehkoupa:


1-exercises/A-array-find/exercise.js
```diff
@@ -17,9 +15,12 @@ let names = [
...
 
-let longNameThatStartsWithA = findLongNameThatStartsWithA(names);
-
-console.log(longNameThatStartsWithA);
+//let longNameThatStartsWithA = findLongNameThatStartsWithA(names);
 
 /* EXPECTED OUTPUT */
 // "Alexandra"
+let findLongNameThatStartsWithA = (namesList) =>
+  namesList.find((name) => name[0] === "A" && name.length > 7);
```
mahri09, line 23: hey, Laleh nice code :), I started to review your code let me give my options here, sometimes we could have long array items and we won't know which of them the long length that's why here I suggest you use the sort() method which is arranging array items from long to short (and reverse) and then find the first long item with find method. like this: function
 findLongNameThatStartsWithA(names) {
  names.sort((a, b) => b.length - a.length);
  let longestName = names.find((name) => name[0] === "A");
  return longestName;
}

1-exercises/I-string-replace/exercise.js
```diff
@@ -13,7 +13,11 @@
 let story =
   "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
 
-let result = story.replace("", "");
+let result = story
+  .replace(/dogs/g, "cats")
```
mahri09, line 17: I liked this code to replace every string dog with a cat,

2-mandatory/1-create-functions.js
```diff
@@ -3,16 +3,13 @@ Write a function that:
...
 
 /*
 Write a function that:
 - Accepts an array as a parameter.
-- Returns a new array containing the same elements, except sorted.
-*/
-function sortArray() {
-}
+- Returns a new array containing the same elements, except sorted.*/
+const sortArray = (arr) => arr.slice().sort();
```
mahri09, line 12: great job using slice() not to modify the previous array an alternative to it you can use [...arr], like this:
function sortArray(array) {
  return [...array].sort();
}

2-mandatory/2-oxygen-levels.js
```diff
@@ -10,8 +10,17 @@
...
+  let newLevel;
+  if (level.substring(level.length - 1) === "%") {
+    newLevel = Number(level.replace("%", ""));
+  } else {
+    return false;
+  }
+  return newLevel > 19.5 && newLevel < 23.5;
+};
+
+const findSafeOxygenLevel = (oxygenLevel) => oxygenLevel.find(isAcceptedLevel);
```
mahri09, line 28: oh good way to use function as a parameter if you wanna see more shorter way you can look at it:
function findSafeOxygenLevel(array) {
  return array.find((item) => {
    let num = item.replace("%", "");
    if (item.includes("%") && num > 19.5 && num < 23.5) {
      return item + "%";
    }
  })[0];
}

2-mandatory/8-password-validator.js
```diff
@@ -22,8 +22,20 @@ Expected Result:
...
 
-function validatePasswords(passwords) {}
+const isValidPassword = (password, index, array) =>
+  password.length >= 5 &&
+  containsUppercaseLetter(password) &&
+  containsLowercaseLetter(password) &&
+  containsNumber(password) &&
+  containsSymbol(password) &&
+  array.findIndex((element) => element === password) === index
+    ? true
```
mahri09: as I know no need to give true or false here (? true: false) function automatically returns you true or false if conditions up will match the given argument!

mahri09 (commented):
@lalehkoupa I really was enjoyed reviewing your code well done! to you and I give some options from my point, thank you!

1-exercises/A-array-find/exercise.js
lalehkoupa: Hi @mahri09 , thanks for reviewing my codes, Great ,it is a good idea when we want to find the longest word start with "A" but in this exercise we supposed to find the first name start with "A" and longer than 7 and the expected output is "Alexandra"

2-mandatory/2-oxygen-levels.js
lalehkoupa: thanks it is great to use includes instead of using substring and comparison, it makes my code much shorter.

lalehkoupa: @mahri09 but I think we don't need to return item+"%" here and you can write your code like this:
function findSafeOxygenLevel(array) {
  return array.find((item) => {
    let num = item.replace("%", "");
    return item.includes("%") && num > 19.5 && num < 23.5;
  });
}

2-mandatory/1-create-functions.js
Awyssa: Indeed great point here Mahri! This is called the Spread operator, and it is used to clone arrays to avoid mutating the original array.
You can read up about it [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) and [here](https://www.programiz.com/javascript/spread-operator).

2-mandatory/1-create-functions.js
```diff
@@ -24,17 +21,22 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
-}
+const modifyStr = (names) => names.trim().replace("/", "").toLowerCase();
+
+const tidyUpString = (arrOfString) => arrOfString.map(modifyStr);
```
Awyssa, line 26: Very nice Laleh! I always like to separate my functions so it is easier to read, modify and understand! If you wanted to combine these two, you could do: 

`const tidyUpString = (arr) => arr.map(e => e.trim().replace("/", "").toLowerCase());`

2-mandatory/1-create-functions.js
```diff
@@ -24,17 +21,22 @@ Write a function that:
...
 
 /*
 Write a function that:
 - Takes an array and an index as input.
 - Returns a new array containing the same elements, but without the element at the passed index.
 */
 
-function remove() {
-}
+const remove = (arr, index) => arr.filter((element) => element !== arr[index]);
```
Awyssa, line 34: Perfect answer! I also really like how you wrote `index` and  `element` as opposed to `i` and `e`. When these functions become 10-20 lines long, readability is going to be super important, and writing code like this is so much better! 🎉

2-mandatory/1-create-functions.js
```diff
@@ -43,9 +45,12 @@ Write a function that:
 - The numbers must be rounded to 2 decimal places.
 - Numbers greater 100 must be replaced with 100.
 */
+const formatPercentage = (arr) => arr.map(formatNumToStr);
```
Awyssa: Again, really nice code Laleh! It works and is super readable! If you wanted to make this more succinct (a fancy word some coders use 😅) you could do this

```
const formatPercentage = (arr) => arr.map(e => e >= 100 ? "100%" : `${Math.round(e * 100) / 100}%`);
```

I believe we only learned about the `conditional / ternary operator` this week, but it's a great tool if we are using only one if statement, aka, we want to check one thing and return a truth value and a false value. 

You can read up about the `ternary operator` [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator).

2-mandatory/2-oxygen-levels.js
Awyssa: Two really good solutions here!!! There is another in-built javascript method/function we can use called `parseFloat()`. This function will automatically replace and trim strings and convert them to `Numbers` for us. You can read up about `parseFloat()` [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat)

This was my solution:
```
const findSafeOxygenLevel = (planets) => {
	const safePlanet = planets.find(planet => {
		if( (planet[planet.length - 1] === "%") && (parseFloat(planet) > 19.5 && parseFloat(planet) < 23.5)) {
			return planet;
		}
	});
	return safePlanet;
};
```

And if you wanted to make this even more succinct you could do:
```
const findSafeOxygenLevel = (planets) => planets.find(planet => (planet[planet.length - 1] === "%") && (parseFloat(planet) > 19.5 && parseFloat(planet) < 23.5) ? planet : false);
```

2-mandatory/3-bush-berries.js
```diff
@@ -21,9 +21,11 @@
...
 */
 
-function isBushSafe(berryArray) {
-  //Write your code here
-}
+const isBushSafe = (berryArray) => {
+  return berryArray.every((value) => value === "pink")
+    ? "Bush is safe to eat from"
+    : "Toxic! Leave bush alone!";
+};
```
Awyssa, line 29: Perfect soltion Laleh!!! This is exactly what I wrote 💃

2-mandatory/4-space-colonies.js
```diff
@@ -14,8 +14,8 @@
         Logic: Only strings that start with A, and finish with family
   
 */
-
-function getSettlers() {}
+const isAllowedToStay = (name) => name.includes("family") && name[0] === "A";
+const getSettlers = (names) => names.filter(isAllowedToStay);
```
Awyssa, line 19: Another perfect answer Laleh! Nice one!!! 🎉

If you wanted to combine these you could do:
`const getSettlers = (names) => names.filter(name => name.includes("family") && name[0] === "A");`

I am sure you knew this already, but just wanted to leave the comment 🙌

2-mandatory/5-eligible-students.js
```diff
@@ -7,7 +7,11 @@
   - Returns an array containing only the names of the who have attended AT LEAST 8 classes
 */
 
-function getEligibleStudents() {}
+const getEligibleStudents = (arrayOfStudent) =>
```
Awyssa, line 13: Love it! Another perfect answer! 🎉

2-mandatory/6-journey-planner.js
```diff
@@ -17,15 +17,12 @@
  Hint: search for string methods like Includes and IndexOf.
 */
 
-function checkCodeIsThere(stringText) {
+const checkCodeIsThere = (stringText) => {
```
Awyssa, line 20: Perfect again! The only very slight suggestion I have is to try and use `const` over `let` if we are not reassigning the value of the variable 🙌

2-mandatory/6-journey-planner.js
```diff
@@ -121,9 +121,10 @@ function getLocationName() {}
 
   Advanced challange: try to use arrow function when invoking an array method.
 */
-function journeyPlanner(locations, transportMode) {
-  // Implement the function body
-}
+const journeyPlanner = (locations, transportMode) =>
```
Awyssa, line 124: These are all so good Laleh! I also *LOVE* how you are chaining your functions (.filter and .map) on different lines. This is a great style of coding, and makes things super readable! My senior developer at my company would offer you a job just on this styling alone! Great work! 🎉 💃 👏

2-mandatory/8-password-validator.js
```diff
@@ -22,8 +22,20 @@ Expected Result:
 PasswordValidationResult=  [false, false, false, false, true]
 
 */
+validatePasswords(["StUFf27%", "Pl3nty!", "Jai33", "shajsaUA**&&", "Pl3nty!"]);
+const validatePasswords = (passwordArray) => passwordArray.map(isValidPassword);
```
Awyssa: You are getting two errors here as you are trying to refer to the variable `validatePasswords` before it has been defined, and trying to call the function `isValidPassword` before it has been defined. 

To fix this you have to place `const isValidPassword =` first, then `const  validatePasswords =`, and then finally call the `function validatePasswords()`. However in order to pass the test you do not have to call the function :)

I do this **all** the time and is one of the most common errors 😅

2-mandatory/8-password-validator.js
```diff
@@ -22,8 +22,20 @@ Expected Result:
...
 
-function validatePasswords(passwords) {}
+const isValidPassword = (password, index, array) =>
```
Awyssa: As this function is quite a few lines, I would encourage you to wrap them in the curly braces / `{ }`. This is just a styling preference, but it makes things much easier to read and easy to understand where the function starts and stops :) 

I would say if a function is over 2 or 3 lines long, you should wrap them in curly braces and use the `return` keyword at the end. 🙌

2-mandatory/8-password-validator.js
Awyssa: This is true! I admit that I used to do this all time but realised it was not necessary only a few months ago 😅

2-mandatory/8-password-validator.js
```diff
@@ -22,8 +22,20 @@ Expected Result:
...
+  password.length >= 5 &&
+  containsUppercaseLetter(password) &&
+  containsLowercaseLetter(password) &&
+  containsNumber(password) &&
+  containsSymbol(password) &&
+  array.findIndex((element) => element === password) === index
+    ? true
+    : false;
+
+// if it passes _all_ checks, return true
```
Awyssa, line 40: Again, amazing code here Laleh! I admit this one took me a while to do as well 😅

I am also super impressed you used and understood the `array` 3rd argument of the [map function ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map)!!!

This was my solution, pretty much the exact same as yours!
```
const validatePasswords = (passwords) => passwords.map((password, index) => {
	return password.length >= 5 &&
			containsUppercaseLetter(password) &&
			containsLowercaseLetter(password) &&
			containsNumber(password) &&
			containsSymbol(password) &&
			!passwords.slice(0, index).includes(password);
});
```

2-mandatory/test.js
```diff
@@ -0,0 +1,11 @@
+PreviousPassword = [
```
Awyssa: I always create a file called test.js to try all my code too!!! 😅
We both must have great minds! 👩‍🔬

2-mandatory/6-journey-planner.js
```diff
@@ -121,9 +121,10 @@ function getLocationName() {}
...
+  locations
+    .filter((element) => isAccessibleByTransportMode(element, transportMode))
+    .map(getLocationName);
```
Awyssa, line 127: These are all so good Laleh! I also LOVE how you are chaining your functions (.filter and .map) on different lines!!! This is a great style of coding, and makes things super readable! My senior developer at my company would offer you a job just on this styling alone! Great work! 🎉 💃 👏

Awyssa (approved):
Great work Laleh! I had a few suggestions on the last one, but overall this is amazing code! Keep up the great work  💃 🎉  🙌

lalehkoupa:
Thanks @Awyssa Michael for reviewing my code and your great suggestions :)

2-mandatory/6-journey-planner.js
lalehkoupa: Sure :)

2-mandatory/test.js
lalehkoupa: Yes It's much easier to manage the code

2-mandatory/6-journey-planner.js
lalehkoupa: oh, thank you so much :)

PR #21, opened 2022-01-20
London-8-JS-Week-4-Busra-Erdogan

BusraElmasErdogan:
null

-----
[View rendered 1-exercises/H-array-methods-2/README.md](https://github.com/BusraElmasErdogan/JavaScript-Core-1-Coursework-Week4-London8/blob/main/1-exercises/H-array-methods-2/README.md)

### DIFF
--- a/1-exercises/C-array-every/exercise.js
+++ b/1-exercises/C-array-every/exercise.js
@@ -5,7 +5,9 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents = group.every((element) => {
+  students.includes(element);
+}); // complete this statement
 
 if (groupIsOnlyStudents) {
   console.log("The group contains only students");
--- a/1-exercises/D-array-filter/exercise.js
+++ b/1-exercises/D-array-filter/exercise.js
@@ -8,7 +8,7 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+let pairsByIndex = pairsByIndexRaw.filter((pair) => pair && pair.length === 2); // Complete this statement
 
 let students = ["Islam", "Lesley", "Harun", "Rukmini"];
 let mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
--- a/1-exercises/F-array-forEach/exercise.js
+++ b/1-exercises/F-array-forEach/exercise.js
@@ -8,7 +8,16 @@
 */
 
 let arr = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
-
+arr.forEach((number) => {
+  if (number % 15 === 0) {
+    return console.log("FizzBuzz");
+  } else if (number % 3 === 0) {
+    return console.log("Fizz");
+  } else if (number % 5 === 0) {
+    return console.log("Buzz");
+  }
+  console.log(number);
+});
 /* EXPECTED OUTPUT */
 
 /*
--- a/1-exercises/I-string-replace/exercise.js
+++ b/1-exercises/I-string-replace/exercise.js
@@ -13,7 +13,11 @@
 let story =
   "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
 
-let result = story.replace("", "");
+let result = story
+  .replace(/dogs/g, "cats")
+  .replace(/day/g, "night")
+  .replace("10", "100000")
+  .replace("great", "brilliant");
 
 /* EXPECTED OUTPUT */
 
--- a/2-mandatory/1-create-functions.js
+++ b/2-mandatory/1-create-functions.js
@@ -3,7 +3,14 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the first five elements of the passed array.
 */
-function first5() {
+function first5(array) {
+  const newArray = [];
+  const numberOfElements = Math.min(array.length, 5);
+  for (let i = 0; i < numberOfElements; i++) {
+    newArray.push(array[i]);
+  }
+  return newArray;
+  // return array.slice(0, 5);
 }
 
 /*
@@ -11,7 +18,8 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(array) {
+  return array.slice().sort();
 }
 
 /*
@@ -24,7 +32,12 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
+
+function tidyUpString(arrayOfStrings) {
+  const newArray = arrayOfStrings.slice();
+  return newArray.map((element) =>
+    element.trim().replace("/", "").toLowerCase()
+  );
 }
 
 /*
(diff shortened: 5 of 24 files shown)


1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,9 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents = group.every((element) => {
```
kolomiets, line 8: Maybe my personal style choice (your may be different and I like the current version!)

I tend to use short parameter names for the lambdas (and I never do this for "proper" named functions):

```javascript
let groupIsOnlyStudents = group.every(e => students.includes(e));
```

The reasoning is that lambdas need to be very simple and short and usually they can be easily understood even with short parameter names. Named functions, in contrast, are typically larger and a good name for parameter really helps.

1-exercises/D-array-filter/exercise.js
```diff
@@ -8,7 +8,7 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+let pairsByIndex = pairsByIndexRaw.filter((pair) => pair && pair.length === 2); // Complete this statement
```
kolomiets, line 11: Take a look at [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

A shorter (and maybe more readable) version would be:

```javascript
let pairsByIndex = pairsByIndexRaw.filter(p => p?.length === 2)
```

1-exercises/F-array-forEach/exercise.js
```diff
@@ -8,7 +8,16 @@
 */
 
 let arr = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
-
+arr.forEach((number) => {
```
kolomiets, line 11: There is a common wisdom (which, as a good programmer, you can go against when you need) - a function should have a single exit point (aka return statement). Having multiple return statements in a function generally makes it more difficult to understand and reason about. Consider this version:

```javascript
arr.forEach((number) => {
  let message = number
  if (number % 15 === 0) {
    message = "FizzBuzz";
  } else if (number % 3 === 0) {
    message = "Fizz";
  } else if (number % 5 === 0) {
    message = "Buzz";
  }
  console.log(message);
});
```

Doing it this way, we also reduce the code repetition (console.log calls) - so, for example, if we want to change the way we output the message, we need to change only one line, not multiple lines as before.

1-exercises/I-string-replace/exercise.js
```diff
@@ -13,7 +13,11 @@
 let story =
   "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
 
-let result = story.replace("", "");
+let result = story
```
kolomiets, line 16: 👍 Call chaining is a very powerful technique - it's may be difficult to grasp, but once you get it, you see it everywhere :)

2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,23 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the first five elements of the passed array.
 */
-function first5() {
+function first5(array) {
```
kolomiets, line 6: It may be a surprise, but this is exactly what [slice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) function does :)
So really it could be as short as:

```javascript
return array.slice(0, 5)
```

2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,23 @@ Write a function that:
...
 }
 
 /*
 Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(array) {
+  return array.slice().sort();
```
kolomiets, line 22: 👍

2-mandatory/1-create-functions.js
```diff
@@ -33,7 +46,10 @@ Write a function that:
 - Returns a new array containing the same elements, but without the element at the passed index.
 */
 
-function remove() {
+function remove(array, index) {
+  const newArray = [...array];
```
kolomiets, line 50: Very good :) What is the difference between this approach and slice() call?

2-mandatory/1-create-functions.js
```diff
@@ -44,7 +60,10 @@ Write a function that:
 - Numbers greater 100 must be replaced with 100.
 */
 
-function formatPercentage() {
+function formatPercentage(arrayOfNumbers) {
```
kolomiets, line 63: Like this approach - an alternative one-liner is probably too wordy:

```javascript
return arrayOfNumbers
    .map(number => Math.round(Math.min(number, 100) * 100) / 100 + "%");
```

2-mandatory/3-bush-berries.js
```diff
@@ -22,7 +22,10 @@
 */
 
 function isBushSafe(berryArray) {
```
kolomiets, line 24: Maybe, just maybe - there is a right or wrong answer here - this may be more readable and easier to support going forward:

```javascript
  const safe = berryArray.every((berry) => berry === "pink")
  return safe 
    ? "Bush is safe to eat from"
    : "Toxic! Leave bush alone!";
}
```

2-mandatory/6-journey-planner.js
```diff
@@ -20,8 +20,8 @@
 function checkCodeIsThere(stringText) {
   let magicWord = "code";
   //edit code below
-  if (stringText) {
-    return stringText;
+  if (stringText.includes(magicWord)) {
```
kolomiets, line 23: This is a perfectly fine solution but it has an interesting property - it performs a search for magicWord twice (you search it when you call `includes` and then again in `indexOf`.

For small strings and infrequently called code, this is absolutely OK. However, in real applications you may work with huge strings or/and your code may be executed thousands of times a second. In this situations, it really matters how efficient your code is.

An alternative implementation that performs search once and have similar readability characteristics:

```javascript
magicWordIndex = stringText.indexOf(magicWord);
if (magicWordIndex !== -1) {
  return magicWordIndex;
} else {
  return "Not found";
}
```

2-mandatory/8-password-validator.js
```diff
@@ -23,7 +23,21 @@ PasswordValidationResult=  [false, false, false, false, true]
 
 */
 
-function validatePasswords(passwords) {}
+function validatePasswords(passwordArray) {
```
kolomiets, line 26:
```javascript
function validatePasswords(passwordArray) {
  return passwordArray.map((password, index, array) => {
    return password.length >= 5 &&
      containsUppercaseLetter(password) &&
      containsLowercaseLetter(password) &&
      containsNumber(password) &&
      containsSymbol(password) &&
      array.findIndex((element) => element === password) === index
  });
}
```
Besides, there is one more thing that's important to mention. When you use '&&' (AND) operator and the first operand is `false`, most programming languages don't need to calculate the second operator, because they know already that the result will be `false`. Consider:

```javascript
false && false == false
false && true == false
false && x == false
```

So it does not matter what `x` is - so JavaScript does not bother and does not perform any calculations for `x`.

It's different for || (OR) operator. JavaScript has to calculate all operands to derive the final answer:

```javascript
false || false == false
false || true == true
false || x == ??? // not sure, need to calculate x first
```

This "feature" has practical implications. For your password validation check, you don't really care if a password has an Uppercase/Lowercase/Number/Symbol if it is shorter than 5 characters. With && approach you can express that and reject the password immediately after the first `falsy` check. With || you have to calculate all conditions to get the final answer.

PR #22, opened 2022-01-20
JavaScript-Core-1-Coursework-Week4-London8-Helena-Rog

hachi-ops:
Not all the tests are passing and not everything is completed.

### DIFF
--- a/1-exercises/C-array-every/exercise.js
+++ b/1-exercises/C-array-every/exercise.js
@@ -5,7 +5,11 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+function isStudent(student,group) {
+  return student && group ? true : false;
+}
+
+let groupIsOnlyStudents = students.every(isStudent); // complete this statement
 
 if (groupIsOnlyStudents) {
   console.log("The group contains only students");
--- a/1-exercises/D-array-filter/exercise.js
+++ b/1-exercises/D-array-filter/exercise.js
@@ -14,7 +14,6 @@ function filterOut(pair){
   }     
 } 
 
-
 let pairsByIndex = pairsByIndexRaw.filter(filterOut); // Complete this statement
 
 let students = ["Islam", "Lesley", "Harun", "Rukmini"];
--- a/1-exercises/D-array-filter/exercise.js
+++ b/1-exercises/D-array-filter/exercise.js
@@ -8,11 +8,11 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-function filterOut(pair){
-  if(pair.length !== 2){
-    console.log(pair)
-  }     
-} 
+function filterOut(pair) {
+  if (pair.length !== 2) {
+    console.log(pair);
+  }
+}
 
 let pairsByIndex = pairsByIndexRaw.filter(filterOut); // Complete this statement
 
--- a/1-exercises/D-array-filter/exercise.js
+++ b/1-exercises/D-array-filter/exercise.js
@@ -8,7 +8,14 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+function filterOut(pair){
+  if(pair.length !== 2){
+    console.log(pair)
+  }     
+} 
+
+
+let pairsByIndex = pairsByIndexRaw.filter(filterOut); // Complete this statement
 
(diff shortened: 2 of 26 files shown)


1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,11 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+function isStudent(student,group) {
+  return student && group ? true : false;
```
jstadnik, line 9: I think this might not work the way you intended. The `&&` operator is for boolean (true/false) values, so instead of checking if the "student" string is in the group (is that what you meant?) it will try to interpret them as true/false values and give you nonsensical answers. If you want to see if the student is in the group, you can use `includes`, e.g. `group.includes(student)`.

1-exercises/F-array-forEach/exercise.js
```diff
@@ -9,6 +9,22 @@
 
 let arr = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
 
+function isFizzOrBuzzOrFizzBuzz(num){
+     if(num%3 === 0 && num%5 === 0){
```
jstadnik, line 13: Great work on this one! :+1:

1-exercises/I-string-replace/exercise.js
```diff
@@ -13,7 +13,7 @@
 let story =
   "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
 
-let result = story.replace("", "");
+let result = story.replace(/dogs/g, "cats").replace("10", "100000").replace("great", "brilliant").replace(/day/g, "night");
```
jstadnik, line 16: Nice work, especially with including the `/g`!

2-mandatory/3-bush-berries.js
```diff
@@ -20,11 +20,26 @@
...
-
-function isBushSafe(berryArray) {
-  //Write your code here
+function hasPink(arr){
+  let pink = "pink";
+  if(arr.includes(pink)){
+    return true;
+  }else{
+    return false;
+  }
```
jstadnik, line 29: The `includes` method returns `true` or `false`, so this can just be `arr.includes(pink)`.

2-mandatory/3-bush-berries.js
```diff
@@ -20,11 +20,26 @@
...
 }
 
+function isBushSafe(berryArray) {  
+    if(berryArray.every(hasPink)){
+      return "Bush is safe to eat from";
+    }else if (berryArray.some(hasPink)){
+      return "Toxic! Leave bush alone!"
+    }else{
+      return "Toxic! Leave bush alone!"
+    }
```
jstadnik, line 39: Nice use of `every` and `some`! That said, the `some` condition has the same result as the final else clause, so it's not technically needed.

2-mandatory/4-space-colonies.js
```diff
@@ -15,7 +15,17 @@
   
 */
 
-function getSettlers() {}
+function isFamilyAndStartWithA(name){
+  if (name.includes("family") && name.startsWith("A")){
+    return true;
+  }else{
+    return false;
+  }
```
jstadnik, line 23: Again, you don't need the `if`/`else` here; it can just be `return (name.includes("family") && name.startsWith("A"))`.

2-mandatory/2-oxygen-levels.js
```diff
@@ -11,7 +11,15 @@
     Some string methods that might help you here are .replace() and .substring().
 */
 
-function findSafeOxygenLevel() {}
+function isLevelSafe(oxygen) {       
+        if (oxygen.includes("%") && oxygen > 19.5 && oxygen < 23.5){
+          return  parseFloat(oxygen);
```
jstadnik: I think this function should return `true` or `false`.

2-mandatory/5-eligible-students.js
```diff
@@ -7,7 +7,17 @@
   - Returns an array containing only the names of the who have attended AT LEAST 8 classes
 */
 
-function getEligibleStudents() {}
+function haveAttendedEnoughClasses(num) {
```
jstadnik: You've got the right idea here, but remember that the list has both names and attendance counts! This parameter isn't just a number, it's a name and a number.

1-exercises/D-array-filter/exercise.js
```diff
@@ -8,7 +8,13 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+function filterOut(pair){
+  if(pair.length !== 2){
+    console.log(pair)
```
jstadnik: A function used for filtering needs to return a boolean.

PR #26, opened 2022-01-21
London Class 8 - Shadi Safaee - JS Core 1 - Week 4

migmow:
except number 8 of mandatory

-----
[View rendered 1-exercises/F-array-forEach/README.md](https://github.com/ShadiSafaee/JavaScript-Core-1-Coursework-Week4-London8/blob/main/1-exercises/F-array-forEach/README.md)

### DIFF
--- a/1-exercises/F-array-forEach/exercise.js
+++ b/1-exercises/F-array-forEach/exercise.js
@@ -9,6 +9,8 @@
 
 let arr = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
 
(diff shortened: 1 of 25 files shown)


1-exercises/F-array-forEach/exercise.js
```diff
@@ -9,6 +9,8 @@
 
 let arr = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
 
+let newArray = arr.forEach(value => console.log(value % 3 === 0 ? "Fizz" : value % 5 === 0 ? "Buzz" : value % 3 === 0 && value % 5 === 0 ? "FizzBuzz" : value));
```
beckyf0keefe, line 12: This isn't giving the correct result for 15, can you see what the issue might be? there is a simple solution to the problem, see if you can spot it

beckyf0keefe (commented):
Looking good overall, just have a look at why the exercise F might not be working, but apart from that really good

PR #73, opened 2022-09-05
WM4 - Kerim Zamir - JS Core 1- Week 4

Arbeits-Sachen:


ozibozkurt (commented):
Good Job Kerim, you are good at coding please keep hard working.

PR #85, opened 2022-09-09
WM4 - Alex Phillip - JS Core 1 - Week 4

Alex-Phillip:


1-exercises/A-array-find/exercise.js
```diff
@@ -5,6 +5,10 @@
 
 // write your code here
 
+function findLongNameThatStartsWithA(names) {
+  return names.find(element => element.length > 7 && element.startsWith("A"))
```
farzad-nazif, line 9: I like how you used && inside of find. Never thought that!

1-exercises/H-array-methods-2/exercise2.js
```diff
@@ -7,7 +7,11 @@
   Tip: use the string method .split() and the array method .join()
 */
 
-function capitalise(str) {}
+function capitalise(str) {
```
farzad-nazif, line 10: You could have used slice as well. But yours is correct as well

1-exercises/I-string-replace/exercise.js
```diff
@@ -13,7 +13,10 @@
 let story =
   "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
 
-let result = story.replace("", "");
+let result = story.replace(/dogs/g, "cats")
+.replace("10", "100000")
```
farzad-nazif, line 17: Do not forget there's replaceAll method as well if there are more of one thing you would like to replace

2-mandatory/1-create-functions.js
```diff
@@ -1,50 +1,81 @@
...
-- Returns a new array containing the same elements, except sorted.
-*/
-function sortArray() {
+// /*
+// Write a function that:
+// - Accepts an array as a parameter.
+// - Returns a new array containing the same elements, except sorted.
+// */
+function sortArray(arr) {
+  let sortedArray = [];
```
farzad-nazif, line 17: You could have used map as well. So you wont need inital variable

farzad-nazif (commented):
Great work Alex, Well done

1-exercises/I-string-replace/exercise.js
Alex-Phillip: Ah thanks Farzad, I didn't know about replaceAll

1-exercises/H-array-methods-2/exercise2.js
Alex-Phillip: Thanks Farzad. It's interesting looking through the code and seeing the different ways people get to the same solution!

Alex-Phillip:
Thanks very much for your comments Farzad, your feedback is appreciated.

PR #116, opened 2022-09-25
NW5 Manchester - Haroon Ali - JavaScript 1 - Week 4

haroon-ali-dev:


2-mandatory/8-password-validator.js
```diff
@@ -23,50 +23,68 @@ PasswordValidationResult=  [false, false, false, false, true]
 
 */
```
baharbazargan62, line 24: Great job Haroon! in this exercise one of the conditions is that Must not be any previous password in the passwords array that i think you forget to consider it.

baharbazargan62 (commented):
Great job Haroon! in this exercise one of the conditions is that Must not be any previous password in the passwords array that i think you forget to consider it.

2-mandatory/8-password-validator.js
haroon-ali-dev: Thank you 😄. I didn't forget it. the isNotRepeated() function does that 😃.

PR #117, opened 2022-09-26
Manchester-NW5-LeilaFaez-Js1-Week4

leilafaez:
Your Details :
Your Name: Leila Faez
Your City: Manchester
Your Slack Name: Leilafaez

Homework Details:
Module: JavaScript-Core-1
Week: week4

### DIFF
--- a/1-exercises/A-array-find/exercise.js
+++ b/1-exercises/A-array-find/exercise.js
@@ -3,7 +3,7 @@
   Using .find(), we'd like to find the first name which starts with A and is longer than 7 letters.
 */
 
-// write your code here
+
 
 let names = [
   "Rakesh",
@@ -17,8 +17,9 @@ let names = [
   "Ahmed",
 ];
 
-let longNameThatStartsWithA = findLongNameThatStartsWithA(names);
 
+let longNameThatStartsWithA =  names.find((element) => element[0] === "A" && element.length > 7);
+ 
 console.log(longNameThatStartsWithA);
 
 /* EXPECTED OUTPUT */
--- a/1-exercises/B-array-some/exercise.js
+++ b/1-exercises/B-array-some/exercise.js
@@ -12,6 +12,11 @@ let pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 // https://nodejs.org/api/process.html#process_process_exit_code
 // process.exit(1);
 
+function isNull(arr){
+  return arr.some(()=>{return arr.includes(null)})
+}
+if(isNull(pairsByIndex)) return;
+
 let students = ["Islam", "Lesley", "Harun", "Rukmini"];
 let mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
 
--- a/1-exercises/C-array-every/exercise.js
+++ b/1-exercises/C-array-every/exercise.js
@@ -5,7 +5,7 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents=group.every((el)=>students.includes(el));
 
 if (groupIsOnlyStudents) {
   console.log("The group contains only students");
--- a/1-exercises/D-array-filter/exercise.js
+++ b/1-exercises/D-array-filter/exercise.js
@@ -8,7 +8,11 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+let pairsByIndex = pairsByIndexRaw.filter((e) => {
+  if (typeof e === "object" && e !== null) {
+    return e.length === 2;
+  }
+}); // Complete this statement
 
 let students = ["Islam", "Lesley", "Harun", "Rukmini"];
 let mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
--- a/1-exercises/E-array-map/exercise.js
+++ b/1-exercises/E-array-map/exercise.js
@@ -3,7 +3,27 @@
 
 let numbers = [0.1, 0.2, 0.3, 0.4, 0.5];
 
-let numbersMultipliedByOneHundred; // complete this statement
+//first
+function multiplied1(num){
+  return num*100;
+}
+let result1 =numbers.map(multiplied1);
+console.log(result1);
+
+//second
+let result2=numbers.map(function multiplied2(num){
+  return num*100;
+});
+console.log(result2);
+
+//third
+let result3=numbers.map(function(num){
+  return num*100;
+});
+console.log(result3);
+
+//forth
+let numbersMultipliedByOneHundred=numbers.map((num)=>num*100) ; // complete this statement
 
 console.log(numbersMultipliedByOneHundred);
 
--- a/1-exercises/F-array-forEach/exercise.js
+++ b/1-exercises/F-array-forEach/exercise.js
@@ -9,6 +9,18 @@
 
 let arr = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
 
+arr.forEach(function (num){
+ if (num % 5 === 0 || num % 3 === 0) {
+    console.log("FizzBuzz");
+  }else if (num % 3 === 0) {
+    console.log ("Fizz");
+  } else if (num % 5 === 0) {
+     console.log("Buzz");
+  
+  }else console.log(num);
+}
+);
+
 /* EXPECTED OUTPUT */
 
 /*
--- a/1-exercises/H-array-methods-2/exercise2.js
+++ b/1-exercises/H-array-methods-2/exercise2.js
@@ -7,14 +7,18 @@
   Tip: use the string method .split() and the array method .join()
 */
 
-function capitalise(str) {}
+function capitalise(str) {
+ const capitalized = str.charAt(0).toUpperCase() + str.slice(1);
+
+ return capitalized;
+}
 
 /* 
   DO NOT EDIT BELOW THIS LINE
   --------------------------- */
-let name = "daniel";
+// let name = "daniel";
 
-console.log(capitalise(name));
+// console.log(capitalise(name));
 console.log(capitalise("hello"));
 
 /* 
--- a/1-exercises/I-string-replace/exercise.js
+++ b/1-exercises/I-string-replace/exercise.js
@@ -13,7 +13,13 @@
 let story =
   "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
 
-let result = story.replace("", "");
+let result = story
+  .replace("dogs", "cats")
+  .replace("day", "night")
+  .replace("10", "100000")
+  .replace("dogs", "cats")
+  .replace("great", "brilliant")
+  .replace("day", "night");
 
 /* EXPECTED OUTPUT */
 
--- a/1-exercises/J-string-substring/exercise2.js
+++ b/1-exercises/J-string-substring/exercise2.js
@@ -14,11 +14,11 @@ let names = [
   "Arron Graham",
 ];
 
-names[0] = names[0].substring();
-names[1] = names[1].substring();
-names[2] = names[2].substring();
-names[3] = names[3].substring();
-names[4] = names[4].substring();
+names[0] = names[0].substring(0,names[0].length-8);
+names[1] = names[1].substring(0, names[1].length - 5);
+names[2] = names[2].substring(0, names[2].length - 8);
+names[3] = names[3].substring(0, names[3].length - 7);
+names[4] = names[4].substring(0, names[4].length - 7);
 
 names.forEach((name) => {
   console.log(name);
--- a/1-exercises/J-string-substring/exercise3.js
+++ b/1-exercises/J-string-substring/exercise3.js
@@ -8,7 +8,8 @@
 
 let statement = "I do not like programming";
 
-let result = "";
+// let result = "I do not like programming".replace("not","");
+let result=statement.substring(0,4)+statement.substring(8);
 
 console.log(result);
 
--- a/2-mandatory/1-create-functions.js
+++ b/2-mandatory/1-create-functions.js
@@ -3,15 +3,20 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the first five elements of the passed array.
 */
-function first5() {
-}
+function first5(array) {
+  const size=5;
+  return array.slice(0,size);
+  
+  }
 
 /*
 Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(arr) {
+  let newArr=arr.slice(0);
+  return newArr.sort();
 }
 
 /*
@@ -24,7 +29,10 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
+function tidyUpString(array) {
+  return array.map((str)=>{
+    return str.trim().replace("/","").toLowerCase();
+  });
 }
 
 /*
@@ -33,7 +41,8 @@ Write a function that:
 - Returns a new array containing the same elements, but without the element at the passed index.
 */
 
-function remove() {
+function remove(array,index) {
+  return array.slice(0,index).concat(array.slice(index+1));
 }
 
 /*
@@ -44,7 +53,14 @@ Write a function that:
 - Numbers greater 100 must be replaced with 100.
 */
 
-function formatPercentage() {
+function formatPercentage(numArr) {
+  return numArr.map((number)=>{
+    if(number>100){
+      return "100%";
+    }else {
+      return parseFloat(number.toFixed(2)) +"%";
+    }
+  });
 }
 
 /* ======= TESTS - DO NOT MODIFY ===== */
--- a/2-mandatory/2-oxygen-levels.js
+++ b/2-mandatory/2-oxygen-levels.js
@@ -11,7 +11,15 @@
     Some string methods that might help you here are .replace() and .substring().
 */
 
-function findSafeOxygenLevel() {}
+function findSafeOxygenLevel(oxLevels) {
+   return oxLevels.find((level) => {
+     return (
+       Number(level.replace("%", "")) > 19.5 &&
+       Number(level.replace("%", "")) < 23.5 &&
+       level.includes("%")
+     );
+   });
+}
 
 /* ======= TESTS - DO NOT MODIFY ===== */
 
--- a/2-mandatory/4-space-colonies.js
+++ b/2-mandatory/4-space-colonies.js
@@ -15,7 +15,15 @@
   
 */
 
-function getSettlers() {}
+function getSettlers(voyagers) {
+  let colonisers=[];
+  voyagers.forEach((name)=>{
+    if(name.includes("family")&& name[0]==="A"){
+      colonisers.push(name);
+    }
+  });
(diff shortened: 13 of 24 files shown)


1-exercises/A-array-find/exercise.js
```diff
@@ -17,8 +17,9 @@ let names = [
   "Ahmed",
 ];
 
-let longNameThatStartsWithA = findLongNameThatStartsWithA(names);
 
+let longNameThatStartsWithA =  names.find((element) => element[0] === "A" && element.length > 7);
```
IanTaiteCYF, line 21: Technically correct. You could consider using the variable "name" instead of the variable "element".

1-exercises/B-array-some/exercise.js
```diff
@@ -12,6 +12,11 @@ let pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 // https://nodejs.org/api/process.html#process_process_exit_code
 // process.exit(1);
 
+function isNull(arr){
+  return arr.some(()=>{return arr.includes(null)})
+}
+if(isNull(pairsByIndex)) return;
+
```
IanTaiteCYF, line 19: I don't think I really understood this exercise however, this is what I did:
```javascript
if (pairsByIndex.some((item) => !item)) {
  process.exit(1);
}
```

1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,7 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents=group.every((el)=>students.includes(el));
```
IanTaiteCYF, line 8: Great.

1-exercises/D-array-filter/exercise.js
```diff
@@ -8,7 +8,11 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+let pairsByIndex = pairsByIndexRaw.filter((e) => {
+  if (typeof e === "object" && e !== null) {
+    return e.length === 2;
+  }
+}); // Complete this statement
```
IanTaiteCYF, line 15: Although it is hard to name things properly, please try to avoid single character variable names.

My version looked like this:
```javascript
let pairsByIndex = pairsByIndexRaw.filter((item) => {
  if (item === undefined) return false;
  if (item === null) return false;
  if (typeof(item) !== "object") return false;
  if (item.length !== 2) return false;
  return true;
});
```
My reason for writing the function like this was to be explicit about the validation rules in the function. Your version is also fine.

1-exercises/E-array-map/exercise.js
```diff
@@ -3,7 +3,27 @@
...
+console.log(result2);
+
+//third
+let result3=numbers.map(function(num){
+  return num*100;
+});
+console.log(result3);
+
+//forth
+let numbersMultipliedByOneHundred=numbers.map((num)=>num*100) ; // complete this statement
```
IanTaiteCYF, line 27: All great. The fourth example is most commonly used.

1-exercises/F-array-forEach/exercise.js
```diff
@@ -9,6 +9,18 @@
 
 let arr = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
 
+arr.forEach(function (num){
+ if (num % 5 === 0 || num % 3 === 0) {
+    console.log("FizzBuzz");
```
IanTaiteCYF, line 14: Not quite right. I think you meant "&&" instead of "||"
```javascript
if (num % 5 === 0 && num % 3 === 0) 
```
I have seen the Fizz Buzz exercise used many times in interviews, and you'd be surprised how many people get it wrong, even senior developers. They mainly get it wrong because they don't play close attention to the rules. Just for fun, this is what my version looked like:
```javascript
arr.forEach((n) => {
  let message = "";
  if (n % 3 === 0) message += "Fizz";
  if (n % 5 === 0) message += "Buzz";
  if (message === "") {
    console.log(n);
  } else {
    console.log(message);
  }
});
```
I strongly recommend that you memorise a correct Fizz Buzz implementation to avoid getting it wrong if you're asked to do it in an interview!

1-exercises/H-array-methods-2/exercise2.js
```diff
@@ -7,14 +7,18 @@
   Tip: use the string method .split() and the array method .join()
 */
 
-function capitalise(str) {}
+function capitalise(str) {
+ const capitalized = str.charAt(0).toUpperCase() + str.slice(1);
+
+ return capitalized;
+}
```
IanTaiteCYF, line 15: Great. Very efficient.

1-exercises/I-string-replace/exercise.js
```diff
@@ -13,7 +13,13 @@
...
   "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
 
-let result = story.replace("", "");
+let result = story
+  .replace("dogs", "cats")
+  .replace("day", "night")
+  .replace("10", "100000")
+  .replace("dogs", "cats")
+  .replace("great", "brilliant")
+  .replace("day", "night");
```
IanTaiteCYF, line 23: Whilst this is technically correct, it's a bit fragile. If the story included another dog, this code would fail. Initially, I thought that the replace method would replaces all occurrences of the first argument, but I learned that's not correct. Here's my version:
```javascript
let result = story
  .replace(/day/g, "night")
  .replace("great", "brilliant")
  .replace(/dogs/g, "cats")
  .replace("10", "100000");
```
The first argument to replace can be a regular expression between two "/" characters. the trailing "g" means match the expression throughout the input string, so any number of "day" and "dogs" occurrences will be correctly replaced.

1-exercises/J-string-substring/exercise2.js
```diff
@@ -14,11 +14,11 @@ let names = [
...
-names[0] = names[0].substring();
-names[1] = names[1].substring();
-names[2] = names[2].substring();
-names[3] = names[3].substring();
-names[4] = names[4].substring();
+names[0] = names[0].substring(0,names[0].length-8);
+names[1] = names[1].substring(0, names[1].length - 5);
+names[2] = names[2].substring(0, names[2].length - 8);
+names[3] = names[3].substring(0, names[3].length - 7);
+names[4] = names[4].substring(0, names[4].length - 7);
```
IanTaiteCYF, line 22: I'm not sure of the purpose of this exercise, but nothing more than this is needed:
```javacript
names[0] = names[0].substring(0, 6);
```

1-exercises/J-string-substring/exercise3.js
```diff
@@ -8,7 +8,8 @@
 
 let statement = "I do not like programming";
 
-let result = "";
+// let result = "I do not like programming".replace("not","");
+let result=statement.substring(0,4)+statement.substring(8);
```
IanTaiteCYF, line 13: Correct. Just watch your whitespace. Check that you format your code with Prettier when you're done. This is more readable code:
```javascript
let result = statement.substring(0, 4) + statement.substring(8);
```

2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,20 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the first five elements of the passed array.
 */
-function first5() {
-}
+function first5(array) {
+  const size=5;
+  return array.slice(0,size);
```
IanTaiteCYF, line 8: Just inline the "5":
```
return arr.slice(0, 5);
```

2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,20 @@ Write a function that:
...
 /*
 Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(arr) {
+  let newArr=arr.slice(0);
+  return newArr.sort();
 }
```
IanTaiteCYF, line 20: Correct. (1) Make a copy of the array. (2) sort it. (3) return the sorted array. This also works and is a bit more "modern":
```
return [...arr].sort();
```
The three dots is called the "spread operator". When used inside "[" and "]", a new array is created, which here is then sorted and returned. It's a very shorthand way of doing the 3 steps above!

2-mandatory/1-create-functions.js
```diff
@@ -24,7 +29,10 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
+function tidyUpString(array) {
+  return array.map((str)=>{
+    return str.trim().replace("/","").toLowerCase();
+  });
 }
```
IanTaiteCYF, line 36: Although the rules for this function state "Removes any forward slashes (/) in the strings.", the data in the test contains only one forward slash, so doing ".replace("/","") makes the test pass, though this form of replace, replaces just the first occurrence of forward slash. If the input string had two or more forward slashes, your code would have failed. You can globally replace a string using this form of replace:
```javascript
str.replace(/\//g, "")
```
Whilst this looks strange, it's unfortunate that regular expressions start and end with the same character (forward slash) that we want to target, so to do so, we have to prefix the forward slash character with a backslash character. This is called "escaping" a character when you prefix it with a backslash. The "g" on the end of the regular expression is a flag used to mean global, i.e. when this match is found anywhere in "str", replace the matched characters with an empty string.

2-mandatory/1-create-functions.js
```diff
@@ -33,7 +41,8 @@ Write a function that:
 - Returns a new array containing the same elements, but without the element at the passed index.
 */
 
-function remove() {
+function remove(array,index) {
+  return array.slice(0,index).concat(array.slice(index+1));
 }
```
IanTaiteCYF, line 46: Correct. You got the "index + 1" part right meaning "the rest of the array". I personally prefer the spread operator:
```javascript
return [ ...arr.slice(0, index), ...arr.slice(index + 1) ];
```

2-mandatory/1-create-functions.js
```diff
@@ -44,7 +53,14 @@ Write a function that:
...
-function formatPercentage() {
+function formatPercentage(numArr) {
+  return numArr.map((number)=>{
+    if(number>100){
+      return "100%";
+    }else {
+      return parseFloat(number.toFixed(2)) +"%";
+    }
+  });
 }
```
IanTaiteCYF, line 64: This is really good. My version was not very readable at all!
```javascript
return `${num > 100 ? "100" : num.toFixed(2).replace(/\.00$/, "").replace(/0$/, "")}%`;
```

2-mandatory/2-oxygen-levels.js
```diff
@@ -11,7 +11,15 @@
...
-function findSafeOxygenLevel() {}
+function findSafeOxygenLevel(oxLevels) {
+   return oxLevels.find((level) => {
+     return (
+       Number(level.replace("%", "")) > 19.5 &&
+       Number(level.replace("%", "")) < 23.5 &&
+       level.includes("%")
+     );
+   });
+}
```
IanTaiteCYF, line 23: This is good clean looking code. I made a couple of small changes which I'll explain below.
```javascript
function findSafeOxygenLevel(oxLevels) {
  const lowerLimit = 19.5;
  const upperLimit = 23.5;
  return oxLevels.find((level) => {
    return (
      level.includes("%") &&
      Number(level.replace("%", "")) > lowerLimit &&
      Number(level.replace("%", "")) < upperLimit
    );
  });
}
```
In large functions numeric literals like "19.5" can be "hidden" in the code. I've created a variable representing these values to make the function more readable.
The predicate function used by ```oxLevels.find``` checks three conditions. When Javascript evaluates expressions that are &&'d together, it evaluates them left-to-right. I've moved the percentage check to be first so that if that expression returns false, the other two expressions are not evaluated. This technique speeds up your function and is sometimes known as "short circuiting" an expression. You could also have used ```.endsWith("%")``` instead of includes.

2-mandatory/4-space-colonies.js
```diff
@@ -15,7 +15,15 @@
...
-function getSettlers() {}
+function getSettlers(voyagers) {
+  let colonisers=[];
+  voyagers.forEach((name)=>{
+    if(name.includes("family")&& name[0]==="A"){
+      colonisers.push(name);
+    }
+  });
+  return colonisers;
+}
```
IanTaiteCYF, line 27: ```voyagers.filter```?

2-mandatory/5-eligible-students.js
```diff
@@ -7,7 +7,9 @@
   - Returns an array containing only the names of the who have attended AT LEAST 8 classes
 */
 
-function getEligibleStudents() {}
+function getEligibleStudents(students) {
+  return students.filter((attend)=>attend[1]>7).map((attend)=>attend[0]);
+}
```
IanTaiteCYF, line 13: Great. Filter chained to map.
The use of numeric literals on line 11 makes me nervous (7 and 0). Could you have introduced a couple of variables here that would make the code self-documenting?

2-mandatory/6-journey-planner.js
```diff
@@ -20,8 +20,8 @@
 function checkCodeIsThere(stringText) {
   let magicWord = "code";
   //edit code below
-  if (stringText) {
-    return stringText;
+  if (stringText.includes(magicWord)) {
+    return stringText.indexOf(magicWord);
```
IanTaiteCYF, line 24: You're searching the string twice with _includes_ and _indexOf_. Consider this alternative way:
```javascript
  let magicWord = "code";
  //edit code below
  const foundAt = stringText.indexOf(magicWord);
  if (foundAt !== -1) {
    return foundAt;
  } else {
    return "Not found";
  }
}
```
Even this example can be condensed however, I prefer more verbose code if it means it's more readable or easier to debug. If _indexOf_ returns _-1_ it means _magicWord_ was not found in _stringText_.

2-mandatory/6-journey-planner.js
```diff
@@ -64,7 +64,9 @@ function checkCodeIsThere(stringText) {
 
   Hint: Use the corresponding array method to split the array.
 */
-function getTransportModes() {}
+function getTransportModes(array) {
+  return array.slice(1,3);
+}
```
IanTaiteCYF, line 69: Whoa, the use of _3_ here looks dodgy. If the input data had 1 location and 5 modes of transport, this would fail. Would this work: ```return array.slice(1, array.length)```?

2-mandatory/6-journey-planner.js
```diff
@@ -81,7 +83,9 @@ function getTransportModes() {}
 
   Hint: Use the corresponding array method to decide if an element is included in an array.
 */
-function isAccessibleByTransportMode() {}
+function isAccessibleByTransportMode(arr,mode) {
+  return arr.includes(`${mode}`);
+}
```
IanTaiteCYF, line 88: Does array have an _includes_ method? I can't remember!
Is using string interpolation (backticks) redundant? Would this do the same thing ```return arr.includes(mode);```?

2-mandatory/6-journey-planner.js
```diff
@@ -123,6 +129,8 @@ function getLocationName() {}
 */
 function journeyPlanner(locations, transportMode) {
   // Implement the function body
+  let result = locations.filter((place) => place.includes(transportMode));
+  return result.map((location) => location[0]);
 }
```
IanTaiteCYF, line 135: I used reduce to solve this, but you definitely have the knack of writing short efficient code. Here's mine:
```javascript
  return locations.reduce(
    (reachableLocations, currentLocationWithTransports) => {
      const [location, ...transportsForLocation] = currentLocationWithTransports;
      if (transportsForLocation.find(mode => mode === transportMode)) {
        return [...reachableLocations, location];
      } else {
        return reachableLocations;
      }
    },
    []
  );
}
```
Mine does the job and is quite fancy, but yours is simpler and a lot more readable than mine, so well done!

2-mandatory/8-password-validator.js
```diff
@@ -23,7 +23,24 @@ PasswordValidationResult=  [false, false, false, false, true]
...
+  //  /\d/g.test(password)&&
+  containsNumber(password)&&
+  //  /[!#$%.*&]/g.test(password)&&
+  containsSymbol(password) &&
+   !prPassword.includes(password)){
+      prPassword.push(password);
+      return true;
+    }else return false;
+  });
+}
```
IanTaiteCYF, line 44: I see you're saving previous passwords. Here's my solution:
```javascript
function validatePasswords(passwords) {
  return passwords.map((password, index) => {
    // - Have at least 5 characters.
    if (password.length < 5) return false;
    // - Have at least one English uppercase letter (A-Z)
    if (!containsUppercaseLetter(password)) return false;
    // - Have at least one English lowercase letter (a-z)
    if (!containsLowercaseLetter(password)) return false;
    // - Have at least one number (0-9)
    if (!containsNumber(password)) return false;
    // - Have at least one non-alphanumeric symbol ("!", "#", "$", "%", ".", "*", "&")
    if (!containsSymbol(password)) return false;
    // - Must not be any previous password in the passwords array.
    if (passwords.slice(0, index).find((pwd) => pwd === password)) return false;
    // Must be OK!
    return true;
  });
}
```
In this version, I'm using the principle of _early exit_ from my _map_ predicate. Rather than string all the password rules together into one giant condition, I have a condition per rule and return _false_ as soon as a rule is broken. This is much easier to debug than a giant condition. I've also left the rule comments in place, as you did too, so I can be sure I have not forgotten a rule.

The checking of the "cannot be a previous password" rule is the hardest part of this exercise. Rather than keep a list of previously processed passwords, I just used slice to get the passwords from the start up to the _current_ password. I didn't test this thoroughly though so it could be a naive implementation.

3-extra/card-validation.js
```diff
@@ -0,0 +1,10 @@
+function onlyNumbers(string) {
+  return /^[0-9]*$/.test(string);
+}
+// check if numbers only / sum > 16 / check if all numbers are same / check if last digit is even
+const ccValidator = (cc) => {
+  const sum = cc.split("").reduce((a, i) => a + parseInt(i, 10), 0);
+  return onlyNumbers(cc) && sum >= 16 && sum / 16 !== parseInt(cc[0], 10) && cc[15] % 2 === 0;
+}
+
+console.log(ccValidator("6666666666666661"));
```
IanTaiteCYF, line 10: Did you find this exercise hard? Just looking through your function, it's pretty hard to tell at a glance what rules are being checked on line 7. Here's mine:
```javascript
function validateCreditCardNumber(cardNumber) {
  // - Number must be 16 digits, all of them must be numbers.
  if (!cardNumber.match(/^\d{16}$/)) return false;
  // - You must have at least two different digits represented (all of the digits cannot be the same).
  if (cardNumber === "".padStart(16, cardNumber[0])) return false;
  // - The final digit must be even.
  if (!cardNumber.match(/[02468]$/)) return false;
  // - The sum of all the digits must be greater than 16.
  if (
    cardNumber.split("").reduce((total, char) => (total += Number(char)), 0) <=
    16
  )
    return false;
  // Must be OK
  return true;
}
```
I left the comments in for the validation rules and used some regular expressions to simplify things.

3-extra/card-validation.js
```diff
@@ -0,0 +1,10 @@
+function onlyNumbers(string) {
+  return /^[0-9]*$/.test(string);
+}
+// check if numbers only / sum > 16 / check if all numbers are same / check if last digit is even
+const ccValidator = (cc) => {
+  const sum = cc.split("").reduce((a, i) => a + parseInt(i, 10), 0);
+  return onlyNumbers(cc) && sum >= 16 && sum / 16 !== parseInt(cc[0], 10) && cc[15] % 2 === 0;
```
haroon-ali-dev, line 7: great job :)

PR #123, opened 2022-09-28
NW5 Manchester - Doris Siu - JavaScript 1 - Week 4

Doris-Siu:


1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,9 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents = group.every((personInGroup) => {
+  return students.some((student) => personInGroup === student);
+}); // complete this statement
```
Nomes27, line 10: great use of the .sort & .every array methods here :) you could also use .includes as an alternative solution within your .every function, for example:
`return students.includes(personInGroup)`

1-exercises/D-array-filter/exercise.js
```diff
@@ -8,7 +8,7 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+let pairsByIndex = pairsByIndexRaw.filter(item => item!== null && typeof item[0] === "number" && typeof item[1] === "number"); // Complete this statement
```
Nomes27, line 11: The filter method works well here

1-exercises/F-array-forEach/exercise.js
```diff
@@ -9,6 +9,20 @@
...
+    else if (num === 5 || num === 10) {
+      console.log("Buzz");}
+      else if (num === 15) {
+        console.log("FizzBuzz")
+      } else {
+        console.log(num);
+      }
+    }
+
+arr.forEach(printing);
```
Nomes27, line 24: your solution works for the array given, well done :) can you think of another way of approaching this which uses the remainder operator? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Remainder
For example, what would happen with your solution above if the array contained numbers beyond 15?

1-exercises/G-array-methods/exercise.js
```diff
@@ -4,7 +4,7 @@
 */
 
 let numbers = [3, 2, 1];
-let sortedNumbers; // complete this statement
+let sortedNumbers = numbers.sort(); // complete this statement
```
Nomes27, line 7: when using .sort() I would create a copy of the array, so that you know there is no mutation happening. for example:
`let sortedNumbers = [...numbers].sort();`
This way you know that  your original numbers array is not going to be mutated, you can read more about the spread operator here https://www.javascripttutorial.net/es6/javascript-spread/

1-exercises/I-string-replace/exercise.js
```diff
@@ -10,10 +10,11 @@
   "I like cats. One night I went to the park and I saw 100000 cats. It was a brilliant night."
 */
 
-let story =
+const story =
   "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
 
-let result = story.replace("", "");
+
+let result = story.replace(/dogs/g,'cats').replace(/day/g,'night').replace(/great/g,'brilliant').replace(/10/g,'100000');
```
Nomes27, line 17: really nice solution and use of regex, well done

2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,18 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the first five elements of the passed array.
 */
-function first5() {
+function first5(array) {
+  return array.slice(0, 5);
```
Nomes27, line 7: this is good, using .slice doesn't mutate the original array :)

2-mandatory/1-create-functions.js
```diff
@@ -24,7 +27,9 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
+function tidyUpString(array) {
+  let tidied = array.map((str) => str.trim().replace("/", "").toLowerCase());
```
Nomes27, line 31: nice chaining of methods here

2-mandatory/1-create-functions.js
```diff
@@ -33,9 +38,14 @@ Write a function that:
 - Returns a new array containing the same elements, but without the element at the passed index.
 */
 
-function remove() {
+function remove(array, index) {
+  let result = [];
+  array.map((x) => {
+    if (array.indexOf(x) != index) {
```
Nomes27, line 44: instead of using an if statement inside a map, can you think of an array method which might be more appropriate here? Remember you can also access the current index within the map function `array.map((x,currentIndex) => {`

2-mandatory/2-oxygen-levels.js
```diff
@@ -11,7 +11,17 @@
...
 
-function findSafeOxygenLevel() {}
+function safeLevel(figure) {
+  let formatted = parseFloat(figure.replace("%",""));
+  return formatted >19.5 && formatted<23.5 && figure.includes("%"); 
+}
+
+
+function findSafeOxygenLevel(array) {
+  let result = array.find(safeLevel);
```
Nomes27, line 21: good use of the .find array method here

2-mandatory/3-bush-berries.js
```diff
@@ -23,6 +23,11 @@
 
 function isBushSafe(berryArray) {
   //Write your code here
+  if (berryArray.every(color => color === "pink")) {
+    return "Bush is safe to eat from" ;
+  } else {
+    return "Toxic! Leave bush alone!";
+  }
 }
```
Nomes27, line 31: this is a really nice concise and readable solution, well done :)

2-mandatory/5-eligible-students.js
```diff
@@ -7,7 +7,14 @@
...
 
-function getEligibleStudents() {}
+function getEligibleStudents(array) {
+  let eligible = array.filter((student) => student[1] >= 8);
+  let eligibleStudents = [];
+  for (let student of eligible) {
+    eligibleStudents.push(student[0]);
+  }
+  return eligibleStudents;
+}
```
Nomes27, line 17: this is good, I really like how you're using your variable names to provide context, it makes your solutions really readable. Instead of a for loop, another approach could be to chain a .map onto your filter to get the first item in each array

2-mandatory/6-journey-planner.js
```diff
@@ -64,7 +64,10 @@ function checkCodeIsThere(stringText) {
 
   Hint: Use the corresponding array method to split the array.
 */
-function getTransportModes() {}
+function getTransportModes(array) {
+  let transportModesArr = array.slice(1, array.length);
```
Nomes27, line 68: the default end of the .slice method is the end of the array, so you can omit the `array.length` here :)

2-mandatory/6-journey-planner.js
```diff
@@ -123,6 +130,10 @@ function getLocationName() {}
 */
 function journeyPlanner(locations, transportMode) {
   // Implement the function body
+  let transportArr = locations
+    .filter((x) => getTransportModes(x).some((y) => y === transportMode))
+    .map((x) => x[0]);
+  return transportArr;
```
Nomes27, line 136: good chaining of array methods here, could you think of alternative variable names instead of x & y?

3-extra/2-sorting-algorithm.js
```diff
@@ -13,9 +13,33 @@ Create a function called sortAges which:
...
+  let right = [];
+
+  for (let i = 1; i < array.length; i++) {
+    array[i] < pivot ? left.push(array[i]) : right.push(array[i]);
+  }
+
+  return quicksort(left).concat(pivot, quicksort(right));
+}
+
+function sortAges(arr) {
```
Nomes27, line 33: really well done on making it onto the extra folder and having passing tests! Just a quick note on .splice, remember it mutates the original array - so always better to make a copy of it first

PR #124, opened 2022-09-29
NW5-Manchester- Yousef Shaikhan- Javascripts1- Week 4

youweb3:


SteveLeicester:
Hi Yousef

Thanks for submitting this pull request.  I can see you have worked on most of the exercises and they are mostly working so well done. But note your find long names beginning with A is slightly wrong. So

```
function findLongNameThatStartsWithA(names){
  return names.find((names) => names[0] === "A" && names.length <= 7);
}
```
should be ( **>= 7 not <= 7**)

```
function findLongNameThatStartsWithA(names){
  return names.find((names) => names[0] === "A" && names.length >= 7);
}

```

Your replace function is missing. You can start with

`  let result = story.replace("10", "100000");`
  
now if you carry on like so...
`result = result.replace("dogs", "cats");`
It may not work because there are two sub(strings) in the story for "dog".  To replace all substrings use this syntax

` result = result.replace(/dogs/gi, "cats");`

It looks like you didn't make the mandatory exercises.  But good effort on these exercises.

Steve

PR #127, opened 2022-09-29
Mohammadreza Nafar -  JavaScript-Core-1-Coursework-Week 4

Mohammadrezanafar:


SteveLeicester:
Thanks for your submission Mohammad.

All the tests pass including the extra tests so a lot of great work.  But one thing to note - passing tests doesn't necessarily mean the code is strictly correct (even if it works).  So I noticed in your soring algorithm (lines 24 and 25)

```
for (let i = 0; i < numArray.length; ++i) {
        if (numArray[i] > numArray[i + 1]) {
```

This will go beyond the end of your array so it should be.  

`for (let i = 0; i < numArray.length - 1; ++i) {`

It just so happens JavaScript is being forgiving so it still works but if you code the algorithm in a different language (C++, C#) it may well crash the program and exit.  

Well that is something to watch for - things may work but possibly for the wrong reason.  I have another example. 
So it looks to me like C-exercise will always give the expected result.  You're comparing different arrays . So on line 6 if you change it to

`let group = ["Austine", "Dany", "Swathi"];`

which now contains only students but as you're joining the two arrays you are comparing `[ "Omar",  "Austine",  "Dany",  "Swathi",  "Lesley",  "Rukmini",  "Austine",  "Dany",  Swathi" ]`

with `[  "Omar",  "Austine",  "Dany",  "Swathi",  "Lesley",  "Rukmini"]`

which again fails giving the group does not contain only students even though it now does. The problem is in the comparison but the exercise was about the every function so we can get a simple boolean ...

```
let groupIsOnlyStudents = group.every(function(groupMember){
   return students.includes(groupMember);
});
```

I-string-replace also fails - it's an easy fix though. This line is good (line16)...

`let result= story.replace(/dogs/gi, "cats");`

giving a new string.  But the next line goes back to the original string (story) and changes it.

`story= story.replace(/day/gi, "night");`

It's just a case of carrying on with the new story...

`result= result.replace(/day/gi, "night");`

So now you preserve the original story and get a new modified story.  


Well that's all I want to say.  It's an excellent submission which is why I'm really digging deeper into the code.  Test driven development is a key component of writing quality code and the more (varied) tests you write the better chance of picking out any faults. But the first line of defence is thinking about each line of code and how it might fail.  It's what makes it a lot of fun (in my opinion anyway)

Hope that is a help.
Steve

Mohammadrezanafar:
Thank you for reviewing my code. Yes, that's correct; I edited my code.

PR #131, opened 2022-09-30
NW5-Manchester- Milad Ebrahimpour - Javascripts1- Week 4

Miladepour:


MustafaAcar-sys (commented):
Good job. At exercise A is your function returning the word which starts with A?

PR #135, opened 2022-09-30
NW5-Kamran_Nikkhah- JavaScript-Core-1-Coursework-Week4

KamNik0044:


SteveLeicester:
Hi Kam

Thanks for your pull request. Good work on the exercises - all working fine.  Some lines of code are missing a **;**  at the end. JavaScript can be quite forgiving - probably more than a team leader when you work as a developer.

I noticed this in your filter exercise. 

```
let numbersMultipliedByOneHundred = numbers.map(multiplyBy100) 
// numbersMultipliedByOneHundred = numbers.map(num => num*100)

```
 Both will work of course but which one is to be preferred?  Any thoughts?
Well it depends is the answer.  If your multiplyBy100 function is something to be used in different places (so a more general utility function) then the former is better.  Otherwise the arrow function version is neat and tidy and does not "pollute" the rest of your code.  That means it is self contained and not dependent on anything else.


One or two failures in the mandatory tests.  Don't think you attempted to formatPercentage but I think the test maybe wrong anyway. It says to two decimal places. So 100% to two places is 100.00%. I can check that though.

Haven't any more time to look at your journey planner solution but that also fails some tests. We can go through that though if you like.

All in all though a strong submission so thanks
Steve

PR #141, opened 2022-09-30
NW5 -Manchester-Adiba Belayete- JS1-WEEK_04

Adibab:


1-exercises/A-array-find/exercise.js
```diff
@@ -9,15 +9,16 @@ let names = [
...
   "Annam",
   "Mikey",
   "Anastasia",
   "Karim",
   "Ahmed",
 ];
 
-let longNameThatStartsWithA = findLongNameThatStartsWithA(names);
+let longNameThatStartsWithA = names.find(
+  (element) => element.length > 7);
```
Nomes27, line 21: this is almost right, good check for the name being longer than 7, but there should also be a check for the name beginning with A :)

1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,10 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+
+let groupIsOnlyStudents = group.every(function(element){
+  return students.includes(element)
+}); // complete this statement
```
Nomes27, line 11: this is really concise, well done

1-exercises/F-array-forEach/exercise.js
```diff
@@ -8,6 +8,28 @@
...
+    if (num % 3 === 0) {
+      console.log("Fizz");
+    } else if (num % 5 === 0) {
+       console.log("Buzz");
+    } else if (num % 3 === 0 && num % 5 === 0) {
+       console.log(FizzBuzz);
+    } else {
+       console.log(num);
+    }
+  })
```
Nomes27, line 21: good use of the remainder operator here, but have a think about the ordering of your if/else statements. When 15 is passed, I would expect this to log Fizzbuzz, however at the moment I would be getting Fizz

1-exercises/G-array-methods/exercise.js
```diff
@@ -4,7 +4,7 @@
 */
 
 let numbers = [3, 2, 1];
-let sortedNumbers; // complete this statement
+let sortedNumbers= numbers.sort(); // complete this statement
```
Nomes27, line 7: `.sort()` mutates the original array, so it's a good idea to create a copy of the array prior to sorting, for example:
`let sortedNumbers= [...numbers].sort()`, there is some more info on the spread operator here https://www.javascripttutorial.net/es6/javascript-spread/

1-exercises/I-string-replace/exercise.js
```diff
@@ -13,7 +13,7 @@
 let story =
   "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
 
-let result = story.replace("", "");
+let result = story.replace("dogs", "cats").replace("day", "night").replace("10", "100000").replace("dogs", "cats").replace("great", "brilliant").replace("day", "night");
```
Nomes27, line 16: good chaining of .replace here. You look to have an extra `day night` replace at the end

2-mandatory/1-create-functions.js
```diff
@@ -33,7 +37,8 @@ Write a function that:
 - Returns a new array containing the same elements, but without the element at the passed index.
 */
 
-function remove() {
+function remove(array, index) {
+  return array.filter((ele, i) => i !== index);
```
Nomes27, line 41: good use of the filter method here, this works well :)

2-mandatory/3-bush-berries.js
```diff
@@ -21,8 +21,16 @@
...
+}
+
 function isBushSafe(berryArray) {
-  //Write your code here
+  if (berryArray.every(toCheckAllBerriesSafe)) {
+    return "Bush is safe to eat from";
+  } else {
+    return "Toxic! Leave bush alone!";
+  }
 }
```
Nomes27, line 34: good use of .every method here. The naming of your param `num` is a bit confusing, something like `colour` might work better, then in your function you can do `return colour === "pink"`

2-mandatory/4-space-colonies.js
```diff
@@ -15,7 +15,12 @@
   
 */
 
-function getSettlers() {}
+function getSettlers(array) {
+  let result = array.filter((num) => {
+    return num.includes("family") && num[0].startsWith("A");
+  });
+  return result;
+}
```
Nomes27, line 23: This is good, using `.filter` here makes sense, only a small thing but I would just think about the variable names you're using sometimes, as it can make your code less readable. For example in the above solution you have used `num`, however each element in the array is actually a string, something like `settlerString` or even `family` might work better :)

2-mandatory/6-journey-planner.js
```diff
@@ -20,8 +20,8 @@
 function checkCodeIsThere(stringText) {
   let magicWord = "code";
   //edit code below
-  if (stringText) {
-    return stringText;
+  if (stringText.indexOf(magicWord)) {
+    return stringText.index;
```
Nomes27, line 24: this is very close, but you are using `.index` and this isn't actually a method on stringText, so it is returning undefined. You want to instead be returning `stringText.indexOf(magicWord)` in your return statement

2-mandatory/7-lane-names.js
```diff
@@ -6,7 +6,11 @@
   HINT: string and array methods that could be helpful (indexOf, filter)
 */
 
-function getLanes() {}
+
+function getLanes(names) {
+  let result = names.filter((array) => array.includes("Lane"));
+  return result;
+}
```
Nomes27, line 13: Good use of .filter and .includes method here well done :) Let me know if you want to go through some of the other Journey Planner questions that you didn't get onto

1-exercises/A-array-find/exercise.js
Adibab: yeah , You are right. I have amended the code. thanks.

1-exercises/F-array-forEach/exercise.js
Adibab: Hi, I have corrected the code. So I have provided comparison operator of (num % 3 === 0 && num % 5 === 0)  first . Now its working. thanks.

1-exercises/G-array-methods/exercise.js
Adibab: yeah. I have amended the code to avoid mutation. thanks.

PR #145, opened 2022-10-04
NW5-Manchester-KhadarMDagal-JS1-W4

khmdagal:
null

-----
[View rendered 1-exercises/F-array-forEach/README.md](https://github.com/khmdagal/JavaScript-Core-1-Coursework-Week4/blob/main/1-exercises/F-array-forEach/README.md)

### DIFF
--- a/1-exercises/A-array-find/exercise.js
+++ b/1-exercises/A-array-find/exercise.js
@@ -21,5 +21,23 @@ let longNameThatStartsWithA = names.find(element => element.startsWith('A') && e
 
 console.log(longNameThatStartsWithA);
 
+const arr = [1,2,3,4,5,6]
+// function smallEnough(a, limit) {
+//   return a.every((element) => (element < limit ? true : false));
+// }
+
+function smallEnough(a, limit) {
+  for (let i = 0; i < a.length; i++) {
+    if (a[i] < limit) {
+      return true;
+    } else {
+      return false;
+    }
+  }
+}
+
+console.log(smallEnough(arr, 0))
+
+
 /* EXPECTED OUTPUT */
 // "Alexandra"
--- a/1-exercises/A-array-find/exercise.js
+++ b/1-exercises/A-array-find/exercise.js
@@ -17,7 +17,7 @@ let names = [
   "Ahmed",
 ];
 
-let longNameThatStartsWithA = findLongNameThatStartsWithA(names);
+let longNameThatStartsWithA = names.find(element => element.startsWith('A') && element.length > 7);
 
 console.log(longNameThatStartsWithA);
 
--- a/1-exercises/B-array-some/exercise.js
+++ b/1-exercises/B-array-some/exercise.js
@@ -15,6 +15,10 @@ let pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 let students = ["Islam", "Lesley", "Harun", "Rukmini"];
 let mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
 
+pairsByIndex.some(function (element) {
+  element === null ? console.log("program exited❎") + process.exit(1) : false;
+});
+
 let pairs = pairsByIndex.map(function (indexes) {
   let student = students[indexes[0]];
   let mentor = mentors[indexes[1]];
--- a/1-exercises/C-array-every/exercise.js
+++ b/1-exercises/C-array-every/exercise.js
@@ -5,7 +5,7 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents = group.every(element => students.includes(element))
 
 if (groupIsOnlyStudents) {
   console.log("The group contains only students");
--- a/1-exercises/D-array-filter/exercise.js
+++ b/1-exercises/D-array-filter/exercise.js
@@ -8,7 +8,10 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+let pairsByIndex = pairsByIndexRaw.filter(
+  (element) => element !== null && element !== false && typeof element !== "string" && element.length === 2
+);
+console.log(pairsByIndex)
 
 let students = ["Islam", "Lesley", "Harun", "Rukmini"];
 let mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
--- a/1-exercises/E-array-map/exercise.js
+++ b/1-exercises/E-array-map/exercise.js
@@ -3,7 +3,7 @@
 
 let numbers = [0.1, 0.2, 0.3, 0.4, 0.5];
 
-let numbersMultipliedByOneHundred; // complete this statement
+let numbersMultipliedByOneHundred = numbers.map(num => num * 100)
 
 console.log(numbersMultipliedByOneHundred);
 
--- a/1-exercises/F-array-forEach/README.md
+++ b/1-exercises/F-array-forEach/README.md
@@ -40,7 +40,7 @@ function formatName(name) {
 }
 
 names.map(formatName).forEach(function(name, index) {
-  console.log(index + ": " + name);
+  console.log(index + ": " + name); 
 });
 ```
 
--- a/1-exercises/F-array-forEach/exercise.js
+++ b/1-exercises/F-array-forEach/exercise.js
@@ -17,42 +17,26 @@ function fuzzFucntion(num) {
       num[i] = "Buzz";
     } else if (num[i] % 3 === 0) {
       num[i] = "Fuzz";
-    } else (
-      num[i]
-    )
+    } else num[i];
   }
   return num;
 }
 
 // console.log(fuzzFucntion(arr));
 const fuzzBuz = [];
- arr.forEach(element => {
+arr.forEach((element) => {
   if (element % 15 === 0) {
-   fuzzBuz.push("FuzzBuzz");
+    fuzzBuz.push("FuzzBuzz");
   } else if (element % 5 === 0) {
     fuzzBuz.push("Buzz");
   } else if (element % 3 === 0) {
     fuzzBuz.push("Fuzz");
-  } else { 
+  } else {
     return fuzzBuz.push(element);
-}
-})
+  }
+});
 
-console.log(fuzzBuz)
-
-/*
- arr.forEach((element, index, array) => {
-  if (array[element] % 15 === 0) {
-    array[element] = "FuzzBuz";
-  } else if (array[element] % 5 === 0) {
-    array[element] = "Buzz";
-  } else if (array[element] % 3 === 0) {
-    array[element] = "Fuzz";
-  } else { 
-  array[element]
-}
-})
-*/
+console.log(fuzzBuz);
 
 /* EXPECTED OUTPUT */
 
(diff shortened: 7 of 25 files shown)


1-exercises/A-array-find/exercise.js
```diff
@@ -17,9 +17,27 @@ let names = [
...
+      return true;
+    } else {
+      return false;
+    }
+  }
+}
+
+console.log(smallEnough(arr, 0))
+
+
```
Ara225, line 41: Bit of surplus code here but very good otherwise!

1-exercises/B-array-some/exercise.js
```diff
@@ -15,6 +15,10 @@ let pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 let students = ["Islam", "Lesley", "Harun", "Rukmini"];
 let mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
 
+pairsByIndex.some(function (element) {
+  element === null ? console.log("program exited❎") + process.exit(1) : false;
```
Ara225, line 19: This definitely works; however, it's not the way .some() would normally be used. It's normally used to answer the question "Does this condition evaluate to true for any elements of the array?" See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some for some usage examples.

Also, good thought to add a console.log statement but, the normal way to execute two statements in a ternary operator like this would be to either call a function or use an if  statement. I've never actually seen + used in this context

1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,7 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents = group.every(element => students.includes(element))
```
Ara225, line 8: 👍

1-exercises/D-array-filter/exercise.js
```diff
@@ -8,7 +8,10 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+let pairsByIndex = pairsByIndexRaw.filter(
+  (element) => element !== null && element !== false && typeof element !== "string" && element.length === 2
```
Ara225, line 12: Solid, might have broken it out into a separate function and/or put the conditions on their own line but that's personal preference

1-exercises/E-array-map/exercise.js
```diff
@@ -3,7 +3,7 @@
 
 let numbers = [0.1, 0.2, 0.3, 0.4, 0.5];
 
-let numbersMultipliedByOneHundred; // complete this statement
+let numbersMultipliedByOneHundred = numbers.map(num => num * 100)
```
Ara225, line 6: 👍

1-exercises/F-array-forEach/README.md
```diff
@@ -40,7 +40,7 @@ function formatName(name) {
 }
 
 names.map(formatName).forEach(function(name, index) {
-  console.log(index + ": " + name);
```
Ara225, line 43: 👍

1-exercises/F-array-forEach/exercise.js
```diff
@@ -9,6 +9,35 @@
...
+    } else if (num[i] % 3 === 0) {
+      num[i] = "Fuzz";
+    } else num[i];
+  }
+  return num;
+}
+
+// console.log(fuzzFucntion(arr));
+const fuzzBuz = [];
+arr.forEach((element) => {
```
Ara225, line 27: I'll take this as an answer however, the way the expected output is written, I think it expects you to do console.log of Fizz/Buzz/number which is more of a normal use for forEach - doing something that has nothing to do with the array, like writing some output to screen or calling a function with each element to do something that doesn't involve changing the element.

1-exercises/G-array-methods/exercise.js
```diff
@@ -4,7 +4,8 @@
 */
 
 let numbers = [3, 2, 1];
-let sortedNumbers; // complete this statement
+let sortedNumbers = numbers.sort()
```
Ara225, line 7: 👍

1-exercises/H-array-methods-2/exercise3.js
```diff
@@ -7,7 +7,7 @@
 let ukNations = ["Scotland", "Wales", "England", "Northern Ireland"];
 
 function isInUK(country) {
-  return; // complete this statement
+  return ukNations.includes(country);
```
Ara225, line 10: 👍

1-exercises/I-string-replace/exercise.js
```diff
@@ -10,10 +10,12 @@
...
 */
 
-let story =
-  "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
+let story = "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
 
-let result = story.replace("", "");
+let result = story.replace(
+  "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.",
+  "I like cats. One night I went to the park and I saw 100000 cats. It was a brilliant night."
```
Ara225, line 17: Out of the box thinking here, most people would have replaced dog with cat, day with night etc, but this is a valid if rigid way of doing it

1-exercises/J-string-substring/exercise.js
```diff
@@ -6,7 +6,7 @@
 
 let statement = "I like programming and dogs";
 
-statement = statement.substring();
+statement = statement.substring(0,statement.length -9);
```
Ara225, line 9: 👍

1-exercises/J-string-substring/exercise2.js
```diff
@@ -21,7 +21,9 @@ names[3] = names[3].substring();
 names[4] = names[4].substring();
 
 names.forEach((name) => {
-  console.log(name);
+  const firstName = name.split(' ')
```
Ara225, line 24: I think it did want you to use substring but this is quite a bit closer to real world good practice :)

1-exercises/J-string-substring/exercise3.js
```diff
@@ -7,8 +7,10 @@
 */
 
 let statement = "I do not like programming";
+let firstPart = statement.substring(0,4);
```
Ara225, line 10: Good work :)

2-mandatory/1-create-functions.js
```diff
@@ -3,17 +3,24 @@ Write a function that:
...
 /*
 Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+let arr = ["a", "n", "c", "e", "z", "f"];
+
+function sortArray(array) {
+return  array.sort();
```
Ara225, line 18: This could be slightly problematic; sort() is strange. It sorts the array in place so the original array will be changed. It's good practice not to change values supplied to a function if you can help it just makes things easier to read at a glance. Also, spec says "Returns a new array" explicitly See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

2-mandatory/1-create-functions.js
```diff
@@ -44,7 +62,10 @@ Write a function that:
 - Numbers greater 100 must be replaced with 100.
 */
 
-function formatPercentage() {
+function formatPercentage(array) {
+  let percentagesSign = "%";
+ return array.map((element) => element > 100 ? '100%' : Math.round((element + Number.EPSILON) * 100) / 100  + percentagesSign)
+  
 }
```
Ara225, line 70: Everything else here checks all the boxes 👍

2-mandatory/2-oxygen-levels.js
```diff
@@ -11,7 +11,13 @@
     Some string methods that might help you here are .replace() and .substring().
 */
 
-function findSafeOxygenLevel() {}
+function findSafeOxygenLevel(array) {
```
Ara225, line 14: So this is sadly not quite right. You're nearly there though! Basically, you need to strip the percentage signs out of both the input and the values you're comparing against. The percentage signs mean that they can't actually be converted to numbers. Javascript then does some funny type stuff instead of what you think it should be doing :) Here are some samples of the strange, non-obvious way this works https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Greater_than#string_to_string_comparison

2-mandatory/3-bush-berries.js
```diff
@@ -23,8 +23,12 @@
 
 function isBushSafe(berryArray) {
   //Write your code here
+  const isSafe = berryArray.every(element => element === "pink");
+   return isSafe ? "Bush is safe to eat from" : "Toxic! Leave bush alone!";
```
Ara225, line 27: Brillant :)

2-mandatory/4-space-colonies.js
```diff
@@ -14,8 +14,31 @@
...
+  "Asimov",
+  "Oscar family",
+  "Avery family",
+  "Archer family",
+  "Just A. family",
+  "A Great family",
+];
+function getSettlers(settlers) {
+  const findFamilies = settlers.filter(
+    (element) => element.includes("family") && element.startsWith("A")
```
Ara225, line 35: Good! All tests pass. The instructions do say "Logic: Only strings that start with A, and finish with family " So, I would say to probaly use endsWith() instead of includes, but it's not important here as the strings that contain family end with family.

2-mandatory/5-eligible-students.js
```diff
@@ -7,7 +7,11 @@
   - Returns an array containing only the names of the who have attended AT LEAST 8 classes
 */
 
-function getEligibleStudents() {}
+function getEligibleStudents(students) {
```
Ara225, line 18: All tests pass :)

2-mandatory/6-journey-planner.js
```diff
@@ -20,11 +20,10 @@
 function checkCodeIsThere(stringText) {
```
Ara225, line 20: The tests in this file have been commented out and it's only partially implimented. Looks good so far though!

2-mandatory/7-lane-names.js
```diff
@@ -6,7 +6,10 @@
   HINT: string and array methods that could be helpful (indexOf, filter)
 */
 
-function getLanes() {}
+function getLanes(array) {
+ return array.filter(elements => elements.includes('Lane'))
```
Ara225, line 10: Good :)

2-mandatory/8-password-validator.js
```diff
@@ -23,13 +23,12 @@ PasswordValidationResult=  [false, false, false, false, true]
 
 */
 
-function validatePasswords(passwords) {}
+function validatePasswords(passwords) {
+  passwords.includes(/[^A-Za-z0-9!#$%.*&]/g) && passwords.length >= 5 ? true : false;
```
Ara225, line 27: The core logic here is good but the function should return an array of results for each element in the input array - it currently returns undefined

Ara225 (commented):
Looks mostly good, a few changes needed :)

3-extra/2-sorting-algorithm.js
```diff
@@ -14,7 +14,17 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
 "think" like a computer and practice your knowledge of basic JavaScript.
 */
 
-function sortAges(arr) {}
+
+function sortAges(arr) {
+  return arr
+    .filter((element) => {
```
Ara225, line 20: Tests pass on this file but instructions do say "removes any non-number data types without using the built-in javascript filter method"

PR #155, opened 2022-12-08
London 9 - turing - susan - JavaScript Core 1 - Week 4

susanssky:


susanssky:
extra:
all parament value become ~~number~~->all parament value become **string**

2-mandatory/2-oxygen-levels.js
```diff
@@ -11,28 +11,35 @@
     Some string methods that might help you here are .replace() and .substring().
 */
 
-function findSafeOxygenLevel() {}
+function findSafeOxygenLevel(arr) {
+  const validNum = arr
+    .filter((x) => x.at(-1) === '%')
+    .map((x) => +x.replace('%', ''))
```
samhoooo, line 17: The unary plus is a very nice way to convert string to number: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus

2-mandatory/2-oxygen-levels.js
```diff
@@ -11,28 +11,35 @@
...
+    .sort((a, b) => a - b)
```
samhoooo, line 18: Is it necessary to do a sorting in here?

2-mandatory/6-journey-planner.js
```diff
@@ -92,7 +97,9 @@ function isAccessibleByTransportMode() {}
    - Returns the name of the location
       e.g: "Tower Bridge"
 */
-function getLocationName() {}
+function getLocationName(arr) {
+  return arr[0];
```
samhoooo, line 101: This is already a good answer in here. 

However, I just like to point out that, in here we assume `[0]` is always exists in `arr`. If `arr` is empty, `undefined` will be returned. There is actually a better (safer) way to get the first element in array: https://stackoverflow.com/a/27180125/12219600

2-mandatory/8-password-validator.js
```diff
@@ -23,7 +23,17 @@ PasswordValidationResult=  [false, false, false, false, true]
...
+function validatePasswords(passwords) {
+  return passwords.map(
+    (x, i, arr) =>
+      x.length >= 5 &&
+      containsUppercaseLetter(x) &&
+      containsLowercaseLetter(x) &&
+      containsNumber(x) &&
+      containsSymbol(x) &&
+      arr.indexOf(x) === i
+  );
```
samhoooo, line 35: ~~We can also use [filter function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) in here~~

Removing this comment as it is not approprate to use filter function in here

3-extra/1-card-vailidator.js
```diff
@@ -0,0 +1,36 @@
+//num.toString()            --->all parament value become number
+//isLengthSixteen           ---against--->Number must be 16 digits
+//isAllNumber               ---against--->invalid characters
+//isMornThanTwoTypeNumber   ---against--->only one type of number
+//isSumGreaterEqualSixteen  ---against--->sum less than 16
+//isLastNumberEven          ---against--->odd final number
+const cardVailidator = (num) => {
+  const n = num.toString();
+  const isLengthSixteen = n.length === 16;
+  const isAllNumber = typeof Number(n) === 'number';
```
samhoooo, line 10: This is not a valid condition to check if the card is all number, because `typeof Number(n)` will always return `'number'`. 

Please refer to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number. Note that `typeof NaN` is also equals to `'number'`

3-extra/2-sorting-algorithm.js
```diff
@@ -14,7 +14,20 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
...
+
+  for (let i = 1; i < filterArr.length; i++) {
+    for (let j = 0; j < i; j++) {
+      if (filterArr[i] < filterArr[j]) {
+        let x = filterArr[i];
+        filterArr[i] = filterArr[j];
+        filterArr[j] = x;
+      }
+    }
+  }
```
samhoooo, line 28: Good answer! 

An extra thing I would like to point out is, this sorting algorithm is called [bubble sort](https://www.geeksforgeeks.org/bubble-sort/). 

You can also [find out](https://www.geeksforgeeks.org/sorting-algorithms/) there are different types of sorting althorithm which can bring optimization in different scenarios: insertion sort, merge sort, quick sort, heap sort... etc

samhoooo (commented):
Good work!
Here are some of my comments and extra knowledge you may want to look into

2-mandatory/2-oxygen-levels.js
susanssky: ....I forgot why I inserted the "sort" method..thank you for reminding...

2-mandatory/6-journey-planner.js
susanssky: yourArray=[undefined,1,2,3,4]
let firstItem = yourArray.find(x=>x!==undefined);
firstItem

these correct?

samhoooo: It would be even safer if you can filter out the `null` values:

```
const yourArray=[undefined,null,1,2,3,4]
const finteredArray = yourArray.filter(x=>x != null);
console.log(finteredArray) // both undefined and null values are filtered out
```

samhoooo: Yes, this is a good way to get first item in array. But it really depends on the data types of your array, and what are you expecting. So in your example:

```
const yourArray=[undefined,1,2,3,4]
const firstItem = yourArray.find(x=>x!==undefined);
console.log(firstItem) // 1
```

firstItem is `1` because we are ignoring `undefined` values. 

If we need to get the "exact" first element, `[0]` is a better approach to go:

 ```
const yourArray=[undefined,1,2,3,4]
const firstItem = yourArray[0]
console.log(firstItem) // undefined
```

PR #158, opened 2022-12-12
London 9 - Lovelace- Bahare-Bahreinian - JavaScript Core 1 - Week 4

Bahare09:


1-exercises/F-array-forEach/README.md
```diff
@@ -54,7 +54,7 @@ Consider this code:
 
 ```js
 let namesFormatted = names.map(format);
-namesFormatted.forEach(log);
+namesFormatted.forEach();
```
mattietea, line 57: These are the docs, don't want to change this

1-exercises/H-array-methods-2/exercise.js
```diff
@@ -15,8 +15,8 @@ let everyone = [
   "Swathi",
 ];
 
-let firstFive; // complete this statement
-let lastFive; // complete this statement
+let firstFive =everyone.slice (0,5)
+let lastFive = everyone .slice (2,7)
```
mattietea: Be sure your spacing is consistent
```suggestion
let firstFive = everyone.slice(0,5)
let lastFive = everyone.slice(2,7)
```

1-exercises/H-array-methods-2/exercise3.js
```diff
@@ -7,7 +7,7 @@
 let ukNations = ["Scotland", "Wales", "England", "Northern Ireland"];
 
 function isInUK(country) {
-  return; // complete this statement
+  return ukNations.includes (country)
```
mattietea: Spacing

```suggestion
  return ukNations.includes(country)
```

1-exercises/I-string-replace/exercise.js
```diff
@@ -13,10 +13,12 @@
 let story =
   "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
 
-let result = story.replace("", "");
-
+let result = story
+              .replaceAll("dog", "cat")
+              .replaceAll("day", "night")
+              .replaceAll("10", "100000")
+              .replaceAll("great", "brilliant");
```
mattietea: This works, is there a better way todo this though? Can you use `split` to turn this into an array and map over the values.  

<details>
<summary>Alternative, don't look until you've tried 😉 </summary>

```js
sentence.split(" ").map((value) => {
  if (value.includes("dog")) {
    return value.replace("dog", "cat");
  } else if (value.includes("day")) {
    return value.replace("day", "night");
  } else if (value.includes("10")) {
    return value.replace("10", "100000");
  } else if (value.includes("great")) {
    return value.replace("great", "brilliant");
  } else {
    return value;
  }
}).join(" ");
```

<details>

2-mandatory/1-create-functions.js
```diff
@@ -44,7 +51,14 @@ Write a function that:
 - Numbers greater 100 must be replaced with 100.
 */
 
-function formatPercentage() {
+function formatPercentage(arr) {
+  return arr.map((item) => {
+    if (item > 100) {
+      return 100 + "%";
+    } else {
+      return Number(item.toFixed(2)) + "%";
```
mattietea, line 58: Because you are doing `+ "%", the `Number` will be turned back into a string - you don't need to cast it.

```suggestion
      return item.toFixed(2) + "%";
```

2-mandatory/2-oxygen-levels.js
```diff
@@ -11,7 +11,15 @@
     Some string methods that might help you here are .replace() and .substring().
 */
 
-function findSafeOxygenLevel() {}
+function findSafeOxygenLevel(arr) {
+  let value = arr
+    .filter((item) => item[item.length - 1] === "%")
+    .map((item) => parseFloat(item))
+    .find((item) => item > 19.5 && item < 23.5);
```
mattietea: While this works, you're looping over the array 3 times. Can you do this problem with only using `find`?

2-mandatory/4-space-colonies.js
```diff
@@ -15,7 +15,12 @@
   
 */
 
-function getSettlers() {}
+function getSettlers(arr) {
+  return arr
+    .map((item) => item.split(" "))
+    .filter((item) => item[0][0][0] === "A" && item[item.length - 1] === "family")    
+    .map((item) => item.join(" "));
+}
```
mattietea: Can you do this, only using `filter`

1-exercises/I-string-replace/exercise.js
Bahare09: i used replace() because the question ask me to use it :) but i will think about split() too

2-mandatory/1-create-functions.js
Bahare09: if i didnt use Number my result is 23.00% however i need 23%

2-mandatory/2-oxygen-levels.js
Bahare09: wow , perfect idea, thank you

Bahare09: Done

2-mandatory/4-space-colonies.js
Bahare09: done , thank you for the great idea :)

1-exercises/I-string-replace/exercise.js
mattietea: You will still want to use `replace`, if you click on "Alternative, don't look until you've tried 😉` it has the solution."

PR #160, opened 2022-12-12
London 9 - Oleh Pysmenko - JS-Core-1 - Week 4 

tiritaki:
Your Details
Your Name: Oleh Pysmenko
Your City: London
Your Slack Name: tiritaki
Homework Details
Module: JS-Core-1
Week: Week 4
Notes
What did you find easy?

What did you find hard?

What do you still not understand?

Any other notes?

### DIFF
--- a/2-mandatory/1-create-functions.js
+++ b/2-mandatory/1-create-functions.js
@@ -3,7 +3,8 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the first five elements of the passed array.
 */
-function first5() {
+function first5(arr) {
+  return arr.slice(0,5)
 }
 
 /*
@@ -11,7 +12,9 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(arr) {
+  let numbersCopy = [...arr]
+  return numbersCopy.sort()
 }
 
 /*
@@ -24,7 +27,12 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
+function tidyUpString(arr) {
+  let newArray = [];
+  for (i in arr) {
+      newArray.push(arr[i].trim().replace('/',"").toLowerCase(0))
+  }
+  return newArray;
 }
 
 /*
@@ -33,7 +41,14 @@ Write a function that:
 - Returns a new array containing the same elements, but without the element at the passed index.
 */
 
-function remove() {
+function remove(arr, ind) {
+  let newarr = []
+  for(let i=0; i<arr.length; i++){
+    if(i!=ind){
+      newarr.push(arr[i])
+    }
+  }
+  return newarr
 }
 
 /*
@@ -44,7 +59,8 @@ Write a function that:
 - Numbers greater 100 must be replaced with 100.
 */
 
-function formatPercentage() {
+function formatPercentage(arr) {
+  return arr.map(num=> num>100 ? '100%' : `${Number((num).toFixed(2))}%`)
 }
 
 /* ======= TESTS - DO NOT MODIFY ===== */
--- a/2-mandatory/1-create-functions.js
+++ b/2-mandatory/1-create-functions.js
@@ -13,8 +13,9 @@ Write a function that:
 - Returns a new array containing the same elements, except sorted.
 */
 function sortArray(arr) {
-  let numbersCopy = [...arr]
-  return numbersCopy.sort()
+  // let numbersCopy = [...arr]
+  // return numbersCopy.sort()
+  return arr.slice().sort()
 }
 
 /*
@@ -28,11 +29,12 @@ Write a function that:
 - Makes the strings all lowercase.
 */
 function tidyUpString(arr) {
-  let newArray = [];
-  for (i in arr) {
-      newArray.push(arr[i].trim().replace('/',"").toLowerCase(0))
-  }
-  return newArray;
+  // let newArray = [];
+  // for (i in arr) {
+  //     newArray.push(arr[i].trim().replace('/',"").toLowerCase(0))
+  // }
+  // return newArray;
+  return arr.map(i=>i.trim().replace('/',"").toLowerCase(0))
 }
 
 /*
@@ -42,13 +44,17 @@ Write a function that:
 */
 
 function remove(arr, ind) {
-  let newarr = []
-  for(let i=0; i<arr.length; i++){
-    if(i!=ind){
-      newarr.push(arr[i])
-    }
-  }
-  return newarr
+  // let newarr = []
+  // for(let i=0; i<arr.length; i++){
+  //   if(i!=ind){
+  //     newarr.push(arr[i])
+  //   }
+  // }
+  // return newarr
+let newArr = arr.slice()
+let removed = newArr.splice(ind,1)
+return newArr
+
 }
 
 /*
--- a/2-mandatory/2-oxygen-levels.js
+++ b/2-mandatory/2-oxygen-levels.js
@@ -10,8 +10,14 @@
 
     Some string methods that might help you here are .replace() and .substring().
 */
+function findSafeOxygenLevel(arr) {
 
-function findSafeOxygenLevel() {}
+  for(let i of arr){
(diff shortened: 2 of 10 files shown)


2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,18 @@ Write a function that:
...
 
 /*
 Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(arr) {
+  let numbersCopy = [...arr]
+  return numbersCopy.sort()
```
nuclear-pasta: when you don't want mutate the original you can use `.slice()`

2-mandatory/1-create-functions.js
```diff
@@ -24,7 +27,12 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
+function tidyUpString(arr) {
+  let newArray = [];
+  for (i in arr) {
+      newArray.push(arr[i].trim().replace('/',"").toLowerCase(0))
+  }
+  return newArray;
```
nuclear-pasta: cool. map would also  work

2-mandatory/1-create-functions.js
```diff
@@ -33,7 +41,14 @@ Write a function that:
...
 
-function remove() {
+function remove(arr, ind) {
+  let newarr = []
+  for(let i=0; i<arr.length; i++){
+    if(i!=ind){
+      newarr.push(arr[i])
+    }
+  }
+  return newarr
```
nuclear-pasta: there is a function that can do this more directly.
```
splice(start)
splice(start, deleteCount)
splice(start, deleteCount, item1)
splice(start, deleteCount, item1, item2, itemN)
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

2-mandatory/2-oxygen-levels.js
```diff
@@ -10,8 +10,14 @@
 
     Some string methods that might help you here are .replace() and .substring().
 */
-
-function findSafeOxygenLevel() {}
+function findSafeOxygenLevel(arr) {
+
+  for(let i of arr){
+    if(i.includes('%') && 19.5< i.replace("%","" ) && i.replace("%","" ) < 23.5){
+        return i
```
nuclear-pasta, line 18: that's nice as you only loop a single time through the array and interrupt as soon as it finds one, but I think the readability could be improved by using some variable name so that it can be easily understood.
What is 19.5? and what is 23.5?
what would happen if I pass this string? "2%0"? The [position](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith) of the % could also be relevant to the data cleaning.

2-mandatory/4-space-colonies.js
```diff
@@ -15,7 +15,10 @@
   
 */
 
-function getSettlers() {}
+function getSettlers(arr) {
+  let A = arr.filter(x=>x[0]=='A')
```
nuclear-pasta: variable name should not start with a capital letter and could be more readable I.E. namesStartingWithA.
Could you use a suggestion made [here](https://github.com/CodeYourFuture/JavaScript-Core-1-Coursework-Week4/pull/160/files#r1046458311)?

2-mandatory/5-eligible-students.js
```diff
@@ -7,7 +7,10 @@
   - Returns an array containing only the names of the who have attended AT LEAST 8 classes
 */
 
-function getEligibleStudents() {}
+function getEligibleStudents(arr) {
+  let eight = arr.filter(x=>x[1]>=8)
```
nuclear-pasta: same here about the variable name. eligibleStudent could suggest already what you are doing.
the number 8 could be included in a variable. So that in the future it can be easier to change it, if the requirements change and the number could change to 7 for example.

2-mandatory/8-password-validator.js
```diff
@@ -23,7 +23,21 @@ PasswordValidationResult=  [false, false, false, false, true]
...
 */
 
-function validatePasswords(passwords) {}
+function validatePasswords(passwords) {
+  let duplicatesArray = passwords.map((e, i, a) => a.indexOf(e) === i)
+  let otherArray = passwords.map(i=>containsUppercaseLetter(i) && containsNumber(i) && containsLowercaseLetter(i)
+  && containsSymbol(i) && i.length>=5)
+  let resultArray=[]
+  for (let j =0; j<passwords.length; j++){
+      if(duplicatesArray[j]===otherArray[j]){
```
nuclear-pasta:
```
if(true){
true}else{
false
}
```
could be simplified because the return value is a bolean.
```
duplicatesArray.map((e, i, a) => e == otherArray[i])
```

3-extra/2-sorting-algorithm.js
```diff
@@ -14,7 +14,15 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
...
 */
 
-function sortAges(arr) {}
+function sortAges(arr) {
+  let nonNumberArr = []
+  for (let item of arr){
+      if (typeof item==='number'){
+        nonNumberArr.push(item)
+      }
+  }
```
nuclear-pasta, line 23: when you create a new array based on a condition you can always use `filter` instead

3-extra/3-card-validator.js
```diff
@@ -0,0 +1,73 @@
...
+// - Make a JavaScript file with a name that describes its contents.
+// - Create a function with a descriptive name which makes it clear what the function does. The function should take one argument, the credit card number to validate.
+// - Write at least 2 comments that explain to others what a line of code is meant to do.
+// - Return a boolean from the function to indicate whether the credit card number is valid.
+// - Use `node` from the command line to test if your code works as expected.
+
+// Good luck!
+
+function cardValidator (cardNumber){
+    let sixDigits = cardNumber.length == 16
```
nuclear-pasta: don't be afraid of using more characters if means being more precise. Because this "sixDigits" even if you mean as a short way of sixteenDigits can be misleading because it means a totally different thing.
Also ,the name sixDigits seem to suggest that the object returned will be something containing six digits, while what you are returning is a boolean. 
Some people like to store the boolean variable in a predictive way: isCarBlue hasChildren 
```
let digitLimit = 16
let hasValidNumberOfDigits =  cardNumber.length == digitLimit
```
but it is not set in stone and you could just go with
```
let digitLimit = 16
if (cardNumber.length !== 16) return false;

```
Since you want ALL the value to be true, as soon as you find a false value you can return the function and not analyze any of the following conditions

2-mandatory/1-create-functions.js
tiritaki: thank you did it with one line and .slice method
return arr.slice().sort()

2-mandatory/1-create-functions.js
tiritaki: did it with one line on map
  return arr.map(i=>i.trim().replace('/',"").toLowerCase(0))

2-mandatory/1-create-functions.js
tiritaki: yes, changed it to newArr.splice(ind,1)

2-mandatory/2-oxygen-levels.js
tiritaki: changed to variables for Oxygen level
added i[4]==="%" to get rid of "2%0"

2-mandatory/4-space-colonies.js
tiritaki: changed it to namesStartingWithA

2-mandatory/5-eligible-students.js
tiritaki: added variable  let numberOfClasses = 8;

2-mandatory/8-password-validator.js
tiritaki: thank you for helping me to shortening it

3-extra/2-sorting-algorithm.js
tiritaki: in task they asked not to use filter :)
removes any non-number data types without using the built-in javascript filter method

3-extra/3-card-validator.js
tiritaki: changed names of variable, left untouched last check, but I totally agree to use on true or false from very beginning

PR #163, opened 2022-12-13
London 9 - Lovelace - Shayan Mahnam - JS Core 1 - Week 4

ShayanMahnam:


2-mandatory/1-create-functions.js
```diff
@@ -33,7 +37,10 @@ Write a function that:
 - Returns a new array containing the same elements, but without the element at the passed index.
 */
 
-function remove() {
+function remove(arr, index) {
```
Zobeir-Rigi, line 40: Hi Shayan, can you explain this code please?

Zobeir-Rigi: I mean the whole function

ShayanMahnam: we have an array, we copy that array to a new array:
`const arr2 = arr.slice();`
or you can write:
`const newArr = arr`
now we want to remove one element from our new array with the index number which we will get from our function, So here we can use splice:
`arr2.splice(index, 1);`

PR #165, opened 2022-12-14
LONDON9-LOVELACE-BOSHRA-MAHMOUDI-JS-CORE1-WEEK4

BoshraM:


Innapoliakova (commented):
In mandatory 6-journey-planner.js in string 68 of your code: I think you deleted the first array of the array instead of the first element of each array. Maybe you should use slice-method to element of each array (not to main array)

PR #171, opened 2022-12-15
London9-Lovelace-Mohamed Abdi-JavaScript-Core-1-Coursework-week4

MohamedAbdi114:


MohamedAbdi114:
@JDysiewicz

1-exercises/A-array-find/exercise.js
```diff
@@ -16,10 +16,18 @@ let names = [
   "Karim",
   "Ahmed",
 ];
+function findLongNameThatStartsWithA(names) {
+	return names.find((name) => name[0] === "A" && name.length > 7);
```
JDysiewicz, line 20: looks good, although what would happen if and empty string was in this list? For this reason, I recommend putting the `name.length` assertion first (as this means you establish that `name[0]` is actually a valid index first).

1-exercises/B-array-some/exercise.js
```diff
@@ -7,11 +7,21 @@
...
 
+function hasNull(index) {
+  return index === null;
+}
+
+let pair = pairsByIndex.some(hasNull);
+if (pair) {
+  return process.exit(1);
+}
+
```
JDysiewicz, line 24: all correct, but as a nitpick, the argument being passed isnt the index of the array (this is the _second_ argument passed when invoking `array.some`), so calling the parameter `index` is a little misleading as the argument being passed is actually the `element`

1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,7 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents = group.every((el)=>students.includes(el)); // complete this statement
```
JDysiewicz, line 8: perfect, although use `const` where you can :P

1-exercises/E-array-map/exercise.js
```diff
@@ -3,7 +3,9 @@
 
 let numbers = [0.1, 0.2, 0.3, 0.4, 0.5];
 
-let numbersMultipliedByOneHundred; // complete this statement
+let numbersMultipliedByOneHundred = numbers.map( function double(number){
+  return number * 100;
+}); // complete this statement
```
JDysiewicz, line 8: this is fine, however for ease of readability I'd recommend either creating your function definition outside of the `.map` and referencing it here (e.g. `numbers.map(multiplyByOneHundred)`) or using an anonymous arrow function (e.g. `numbers.map((x) => x*100)`).

1-exercises/F-array-forEach/exercise.js
```diff
@@ -8,7 +8,17 @@
...
+  let result = "";
 
+  if(number % 3 === 0){ result += `'Fizz'`}
+  if(number % 5 === 0){ result += `'Buzz'`}
+
+
+  if(result === "") {result = number;}
+
+  console.log(result);
+})
```
JDysiewicz, line 21: same comment as above; this is fine but declare your function outside the scope of the `.forEach` and reference it, or use a arrow function.

2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,17 @@ Write a function that:
...
+ }
 
 /*
 Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(arr) {
+  return arr.slice().sort();
```
JDysiewicz, line 16: Something to be wary of is that the javascript `.sort()` assumes things are strings, so it will sort `[2,3,1,11, 23]` as [1,11,2,23,3]`. For this reason, it's best to pass a comparision function in.  https://www.w3schools.com/js/js_array_sort.asp

2-mandatory/1-create-functions.js
```diff
@@ -24,16 +26,22 @@ Write a function that:
...
 /*
 Write a function that:
 - Takes an array and an index as input.
 - Returns a new array containing the same elements, but without the element at the passed index.
 */
 
-function remove() {
+function remove(arr, index) {
+  let newArray = arr.slice();
+  newArray.splice(index, 1);
```
JDysiewicz, line 43: This will work, however it's neater to try and use the array methods `map, filter. reduce` whereever possible. In this case, this could be done with `.filter` via
```
function remove(arr, index) {
   return arr.filter((elem, idx) => idx !== index)
}
```

2-mandatory/3-bush-berries.js
```diff
@@ -23,7 +23,12 @@
 
 function isBushSafe(berryArray) {
   //Write your code here
+  if(berryArray.every((berry) => berry === "pink")) {
+    return "Bush is safe to eat from";
 }
+    return "Toxic! Leave bush alone!";
```
JDysiewicz, line 29: nice use of an early return to remove the need for an `else` statement!

JDysiewicz (commented):
looks good, a couple things around using native array methods map/filter/reduce in a neater way, but all looks correct good job!

PR #173, opened 2022-12-15
Js-core1-week4-Ebrahim Beiati-Asl

ebrahimbeiati:
I did this homework , i would be happy to see your review

### DIFF
--- a/1-exercises/B-array-some/exercise.js
+++ b/1-exercises/B-array-some/exercise.js
@@ -11,14 +11,22 @@ let pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 // If there is a null value in the array exit the program with the error code
 // https://nodejs.org/api/process.html#process_process_exit_code
 // process.exit(1);
+ function isNull(element) {
+   return element === null;
+ }
 
 let students = ["Islam", "Lesley", "Harun", "Rukmini"];
 let mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
 
 let pairs = pairsByIndex.map(function (indexes) {
+
+ if (pairsByIndex.some(isNull)) {
+   process.exit(1);
+ }
+  
   let student = students[indexes[0]];
   let mentor = mentors[indexes[1]];
   return [student, mentor];
-});
+})
 
 console.log(pairs);
--- a/1-exercises/C-array-every/exercise.js
+++ b/1-exercises/C-array-every/exercise.js
@@ -5,7 +5,10 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents ;
+let student = group.map((student) => students.includes(student));
+
+// complete this statement
 
 if (groupIsOnlyStudents) {
   console.log("The group contains only students");
--- a/1-exercises/D-array-filter/exercise.js
+++ b/1-exercises/D-array-filter/exercise.js
@@ -8,7 +8,10 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+let pairsByIndex = pairsByIndexRaw.filter((item) => !!item && item.length === 2); 
+
+
+// Complete this statement
 
 let students = ["Islam", "Lesley", "Harun", "Rukmini"];
 let mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
--- a/1-exercises/G-array-methods/exercise.js
+++ b/1-exercises/G-array-methods/exercise.js
@@ -4,7 +4,7 @@
 */
 
 let numbers = [3, 2, 1];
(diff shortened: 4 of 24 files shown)


1-exercises/B-array-some/exercise.js
```diff
@@ -11,14 +11,22 @@ let pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 // If there is a null value in the array exit the program with the error code
 // https://nodejs.org/api/process.html#process_process_exit_code
 // process.exit(1);
+ function isNull(element) {
+   return element === null;
+ }
```
JDysiewicz, line 16: good choice of naming convention; functions that return booleans often are named `is...`

1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,10 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents ;
+let student = group.map((student) => students.includes(student));
```
JDysiewicz, line 9: small thing that won't cause a programming issue but can be visually harder to read; you're using the same variable name `student` to refer to two different things in this line. It works because it's a different scope, but try and avoid [shadowing](https://www.geeksforgeeks.org/variable-shadowing-in-javascript/) where possible

1-exercises/D-array-filter/exercise.js
```diff
@@ -8,7 +8,10 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+let pairsByIndex = pairsByIndexRaw.filter((item) => !!item && item.length === 2);
```
JDysiewicz, line 11: `!!` is nice shorthand for checking for a variables existence, however be wary this the `!` operator will actually check for [Falsey](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) values when used this way. This means that, if `0` is a valid value, `!0` evaluates to `true` as `0` is a falsey value.

1-exercises/G-array-methods/exercise.js
```diff
@@ -4,7 +4,7 @@
 */
 
 let numbers = [3, 2, 1];
-let sortedNumbers; // complete this statement
+let sortedNumbers= numbers.sort(); // complete this statement
```
JDysiewicz, line 7: this will work in this case, however the `.sort()` method in javascript is silly in that it treats all values as strings all the time, meaning that `[1,3,11,23].sort() = [1,11,23,3]` - so make sure to always pass a comparison function https://www.w3schools.com/js/js_array_sort.asp

2-mandatory/1-create-functions.js
```diff
@@ -1,17 +1,21 @@
+const { StrictMode } = require("react/cjs/react.production.min");
```
JDysiewicz, line 1: you don't need this :P

2-mandatory/1-create-functions.js
```diff
@@ -1,17 +1,21 @@
...
 }
 
 /*
 Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(arr) {
+  return arr.slice().sort();
```
JDysiewicz, line 18: remember to use a comparison function with array.sort

2-mandatory/1-create-functions.js
```diff
@@ -24,7 +28,8 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
+function tidyUpString(arr) {
+  return arr.map((String) => String.trim().replace("/","").toLowerCase());
```
JDysiewicz, line 32: careful of your naming here; variables in javascript are `lowerCamelCase` by convention, and the `String` name is actually a key word for the `String` class in javascript so should not be used as a variable name.

2-mandatory/1-create-functions.js
```diff
@@ -44,7 +52,11 @@ Write a function that:
 - Numbers greater 100 must be replaced with 100.
 */
 
-function formatPercentage() {
+function formatPercentage(arr) {
+    return arr
+      .map((x) => (x > 100 ? "100" : x.toFixed(2)))
+      .map((y) => `${+y}%`);
```
JDysiewicz, line 58: nice use of a ternary here, although you can do both of these `.map` operations in a single `.map` to save going through the array twice; is more efficient and easier to read

2-mandatory/3-bush-berries.js
```diff
@@ -22,6 +22,10 @@
 */
 
 function isBushSafe(berryArray) {
+    return berryArray.every((element) => element === "pink")
+      ? `Bush is safe to eat from`
+      : `Toxic! Leave bush alone!`;
```
JDysiewicz, line 27: nice use of a ternary!

JDysiewicz (commented):
some really solid work here @ebrahimbeiati , you should be proud! :+1:

PR #174, opened 2022-12-15
London9-Marziyeh -Azhdari-JavaScript-Core-1-Coursework-Week4

MariAzhdari:


khmdagal (commented):
Hi Marziyeh-Azhdari  you have done very good work here. What I would recomend is commit more often.

1-exercises/A-array-find/exercise.js
```diff
@@ -17,6 +17,11 @@ let names = [
   "Ahmed",
 ];
 
+function findLongNameThatStartsWithA(names) {
+  let nameArray = names.find((element) => element.startsWith("A") && element.length > 7);
```
mcarballopacheco, line 21: Great use of arrow functions!

1-exercises/H-array-methods-2/exercise.js
```diff
@@ -14,9 +14,8 @@ let everyone = [
   "Austine",
   "Swathi",
 ];
-
-let firstFive; // complete this statement
-let lastFive; // complete this statement
+let firstFive = everyone.slice(0, 5); // complete this statement
+let lastFive = everyone.slice(2, 7);
```
mcarballopacheco, line 18: This would work for this case, but if you start adding new names to everyone then it would fail. You could also do a more general solution like:

`let lastFive = everyone.slice(-5);`

2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,17 @@ Write a function that:
...
 }
 
 /*
 Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(arr) {
+  return arr.slice().sort();
```
mcarballopacheco, line 16: Very nice way to create a copy of an array using slice().

Actually making copies when programming is not always as easy as one may think. If you want, you could have a look at this discussion about how to copy things in javascript:
https://www.freecodecamp.org/news/copying-stuff-in-javascript-how-to-differentiate-between-deep-and-shallow-copies-b6d8c1ef09cd/

2-mandatory/4-space-colonies.js
```diff
@@ -15,7 +15,18 @@
   
 */
 
-function getSettlers() {}
+function getSettlers() {
+  function getSettlers(arr) {
+    let newArr =  [];
+  for(let el of arr){
+if (el.startsWith("A") && el.includes("family")){
```
mcarballopacheco, line 22: For family, you could have used .endsWith("family"). If not a string with "family" in the middle of the string would still pass this test.

2-mandatory/8-password-validator.js
```diff
@@ -23,7 +23,19 @@ PasswordValidationResult=  [false, false, false, false, true]
...
-function validatePasswords(passwords) {}
+function validatePasswords(passwords) {
+  return passwords.map(
+    (element, i, arr) =>
+      element.length >= 5 &&
+      containsLowercaseLetter(element) &&
+      containsNumber(element) &&
+      containsUppercaseLetter(element) &&
+      containsSymbol(element) &&
+      arr.indexOf(element) === i
```
mcarballopacheco, line 34: Great use of the .indexOf method Mari!

mcarballopacheco (commented):
Great work, Mari!

PR #175, opened 2022-12-15
London9-Lovelace-Oleksii Chepurnyi-JavaScript-Core-1-Coursework-Week4

OleksChep:


1-exercises/D-array-filter/exercise.js
```diff
@@ -4,11 +4,17 @@
   It is decided that array items which are not pairs should be filtered out
   - Finish the statement on line 11 to produce an array with valid content
   - Do not edit any of the existing code
+  ------------Вам предоставляется программа, которая регистрирует пары между наставниками и учениками.
```
maxf, line 7: Again, be careful with what you commit before you commit it. Using `git diff` before committing should help making sure you're committing only what's necessary

maxf (commented):
Very good. Looks like you're really getting there and ready for the next course.
But you really need to make an effort with:
- your git commits. You have too many and the log is not helpful 
- the tidiness of your code. Indentation is often wrong, there's some white space missing or extra. Some comments should be removed, as well as some unnecessary `console.log`. You really need to pay attention to details.

PR #176, opened 2022-12-16
London 9 - Lovelace - Howard Sun - JavaScript-core1 - Week4 

howard-ss:


OlushoreOdude:
hi Howard, great use of functions.

1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,12 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents = group.every(isIncluded ) // complete this statement
+   function isIncluded (group){
+        for (let i=0 ; i < group.length; i++){
+           return students.includes('group[i]');
```
mcarballopacheco: This would not pass when the group only has students. There are a couple of errors, first in the line `return students.includes('group[i]');` group should not be given as an array as group is the input to the function. But also, the input to the function here is each value inside of the array therefore you do not need a for loop in the function isIncluded.

One thing I wanted to mention. In this case, the test passes because the only example is false and the function return false, but it is important that you check that function also has the right functionality when you expect it to be true. For example, you can change the line 6 to `let group = ["Austine", "Dany", "Swathi"];` and check what happens then (it should be true but with your function it will still be false).

howard-ss: Hi, Martin
You're absolutely right !
I had double checked it and re-wrote this function as following and it returns the opposite when  I deleted ''Daniel" from group array:
 **let groupIsOnlyStudents = group.every(x=>students.includes(x))** 
 
 Thanks!
 Howard

1-exercises/F-array-forEach/exercise.js
```diff
@@ -9,6 +9,25 @@
...
+        for(let index=1; index<=5;index++){
+          arr.splice((index*3-1),1,`'Fizz'`);
+        } ;
+        for(let index=2; index<=3;index++){
+          arr.splice((index*index),1,`'Buzz'`);
+       };
+          arr.splice((arr.length-1),1,`'FizzBuzz'`);
+        
+       console.log(name)
+  })
```
mcarballopacheco, line 22: While this function works for the given array, if the array includes an extra number, e.g., up to 16 then it would print FizzBuzz for 16 instead of for 15 which is not the expected behaviour. I think it would be simpler to write the function only as a function of index and then just write a number of if statements to check if the index is multiple of 3, 5 or both and then print the right option for each case. In that way, the function would be more general.

1-exercises/H-array-methods-2/exercise2.js
```diff
@@ -1,21 +1,27 @@
 /*
   Array methods - .join()
   -------------------------
-  Complete the capitalise function 
+  Complete the capitalize function
```
mcarballopacheco, line 4: This is a very minor comment, but there is nothing wrong with the spelling of capitalise. Capitalise is british english while capitalize is american english.

1-exercises/H-array-methods-2/exercise2.js
```diff
@@ -1,21 +1,27 @@
...
   It should return a string with the first letter in uppercase
-  For example, capitailise("hello") should return "Hello"
+  For example, capitalize("hello") should return "Hello"
   Tip: use the string method .split() and the array method .join()
 */
 
-function capitalise(str) {}
+function capitalize(str) {
+    
+  return str.split('')[0].toUpperCase()+str.slice(1);
```
mcarballopacheco, line 12: Could you do this using the array method .join()?

1-exercises/I-string-replace/exercise.js
```diff
@@ -13,7 +13,12 @@
 let story =
   "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
 
-let result = story.replace("", "");
+let result1 = story.replace('great', "brilliant");
+let result2 = result1.replace(10,100000);
+let result3 = result2.replace(/day/gi,'night');
```
mcarballopacheco, line 18: This could have worked even if it was a regular expression and it would probably be easier. For example:

`let result3 = result2.replace('day','night');`

2-mandatory/1-create-functions.js
```diff
@@ -3,17 +3,19 @@ Write a function that:
...
+  }
 
 /*
 Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(arr) {
+     return arr.slice().sort();
```
mcarballopacheco, line 16: Very nice way to create a copy of an array using slice().

Actually making copies when programming is not always as easy as one may think. If you want, you could have a look at this discussion about how to copy things in javascript:
https://www.freecodecamp.org/news/copying-stuff-in-javascript-how-to-differentiate-between-deep-and-shallow-copies-b6d8c1ef09cd/

2-mandatory/2-oxygen-levels.js
```diff
@@ -1,17 +1,27 @@
...
 
     Some string methods that might help you here are .replace() and .substring().
 */
 
-function findSafeOxygenLevel() {}
+function findSafeOxygenLevel(oxygenLevel) {
+   let safeLevel = oxygenLevel
+                   .filter ((a)=>a.at(-1)==='%')
+                   .map((b)=> + b.replace('%',''))
+                   .sort((a,b) => a-b)
```
mcarballopacheco, line 21: I don't think you need to sort the array here. Because you are looking for the first planet. For example, in this array `["24.2%", "11.3%", "19.9%", "23.1%", "29.3%", "19.8%"]`, you would get 19.8% from your function, but you should get 19.9%.

2-mandatory/3-bush-berries.js
```diff
@@ -10,19 +11,27 @@
...
 
 function isBushSafe(berryArray) {
   //Write your code here
+  // Loop through the array of berries
+  for (const berry of berryArray) {
+    // If any berry is not pink, return "Toxic! Leave bush alone!"
+    if (berry !== "pink") return "Toxic! Leave bush alone!";
+  }
+  // If all berries are pink, return "Bush is safe to eat from"
+  return "Bush is safe to eat from";
```
mcarballopacheco, line 34: This would work. But could you do this with .some() or .every()?

2-mandatory/5-eligible-students.js
```diff
@@ -2,12 +2,27 @@
...
-function getEligibleStudents() {}
+function getEligibleStudents(arrays) {
+    //create an empty array to store the arr who will sit an exam
+     const sitExamName = [];
+     //loop through the array of arrays
+     for ( arr of arrays ){
+    // check if the array includes a name with 8 attendance counts
+       if(arr[1] >= 8){
+        //add the arr to the sitExamName array
+        sitExamName.push(arr)
```
mcarballopacheco, line 20: You could have already done: `sitExamName.push(arr[0])` and then you would not need to use map in line 24.

2-mandatory/8-password-validator.js
```diff
@@ -23,7 +23,18 @@ PasswordValidationResult=  [false, false, false, false, true]
...
-function validatePasswords(passwords) {}
+function validatePasswords(passwords) {
+   return passwords.map(
+     (x, i, arr) =>
+       x.length >= 5 &&
+       containsUppercaseLetter(x) &&
+       containsLowercaseLetter(x) &&
+       containsNumber(x) &&
+       containsSymbol(x) &&
+       arr.indexOf(x) === i
```
mcarballopacheco, line 34: Great use of the .indexOf method Howard!

mcarballopacheco (commented):
Great work, Howard!

1-exercises/C-array-every/exercise.js
mcarballopacheco: Yes, that works!

PR #181, opened 2022-12-16
London 9 - Lovelace - Yigit Senoglu - JS-Core-1 - Week 4

ys35code:


1-exercises/A-array-find/exercise.js
```diff
@@ -17,6 +17,10 @@ let names = [
   "Ahmed",
 ];
 
+function findLongNameThatStartsWithA(names) {
+  return names.find((name) => name[0] === "A" && name.length > 7);
+}
```
ebrahimbeiati, line 22: Well done

1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,8 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents;
```
maxf, line 8: This variable is never assigned any value.

1-exercises/D-array-filter/exercise.js
```diff
@@ -8,7 +8,9 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+let pairsByIndex = pairsByIndexRaw.filter(
+  (val) => Array.isArray(val) && val.length === 2
```
maxf, line 12: Small detail, when an arrow function only has one argument you don't need the brackets: `...filter(val => Array(...`

1-exercises/E-array-map/exercise.js
```diff
@@ -3,11 +3,15 @@
 
 let numbers = [0.1, 0.2, 0.3, 0.4, 0.5];
 
-let numbersMultipliedByOneHundred; // complete this statement
+// let numbersMultipliedByOneHundred = numbers.map((number) => number * 100);
+
+let numbersMultipliedByOneHundred = numbers.map(function (number) {
```
maxf, line 8: It could be a `const` instead of a `let`, by the way. It's better to make sure it's not mistakenly changed later if you don't indent to.

1-exercises/H-array-methods-2/exercise2.js
```diff
@@ -7,7 +7,13 @@
...
 */
 
-function capitalise(str) {}
+function capitalise(str) {
+  const words = str.split(" ");
+  for (let i = 0; i < words.length; i++) {
+    words[i] = words[i].charAt(0).toUpperCase() + words[i].slice(1);
+  }
+  return words.join(" ");
+}
```
maxf, line 16: Well done. Now you should try to rewrite it without using a loop but using `map()` instead. The code will look much better.

1-exercises/J-string-substring/exercise3.js
```diff
@@ -8,7 +8,9 @@
 
 let statement = "I do not like programming";
 
-let result = "";
+let result =
+  statement.substring(0, statement.indexOf("not")) +
```
maxf, line 12: very good to use `indexOf`

2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,20 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the first five elements of the passed array.
 */
-function first5() {
+function first5(arr) {
+  const first5Elements = arr.slice(0, 5);
+  return first5Elements;
```
maxf, line 8: you can of course spare the `const` and just write `return arr.slice(0, 5)`

2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,20 @@ Write a function that:
...
 /*
 Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(arr) {
+  const newArray = arr.slice();
+  newArray.sort((a, b) => a - b);
+  return newArray;
```
maxf, line 19: Again, not mandatory but you can write `return arr.slice().sort(...)`. You don't want to do it too much generally if your code is going to become too complex to read, but in this case the result is simple enough.

2-mandatory/1-create-functions.js
```diff
@@ -24,7 +29,13 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
+function tidyUpString(arr) {
+  for (let i = 0; i < arr.length; i++) {
```
maxf, line 33: using `.map` will also help you make your code simpler

2-mandatory/1-create-functions.js
```diff
@@ -44,7 +57,19 @@ Write a function that:
 - Numbers greater 100 must be replaced with 100.
 */
 
-function formatPercentage() {
+function formatPercentage(numbers) {
+  const formattedNumbers = [];
+
+  for (const number of numbers) {
```
maxf, line 63: Try again, using `map()`.
Generally, when you need to convert an array to another array with the same number of elements, you should use `map`

2-mandatory/6-journey-planner.js
```diff
@@ -64,7 +63,10 @@ function checkCodeIsThere(stringText) {
 
   Hint: Use the corresponding array method to split the array.
 */
-function getTransportModes() {}
+function getTransportModes(locationInfo) {
+  const location = locationInfo.shift();
+  return locationInfo;
```
maxf, line 68: I think you want to return `location`

maxf (commented):
Excellent. Very good, apart from a couple of places.
Especially impressed with the clarity of your github history and the tidiness of your code. Well done

PR #186, opened 2022-12-16
London-9-Natalie Zablotska-JavaScript-Core1-Coursework-Week-4

nataliiazab:


khmdagal (commented):
Hi Natalia, you have done very good work, you commited more often which is very good, and keep that one. what I would recommend is when you commiting take the comments out first and then commite after that you can bring back you comments int the VS code you can remember what you want do later.  For the mandatory part 6 and 8 do you able to reach you tech bady and ask if you can get help, or the main slack channel. Finally step 5 you have said 'soved with help' this is good one you have said that, which show you appricated and learning, well done

2-mandatory/7-lane-names.js
```diff
@@ -6,7 +6,10 @@
   HINT: string and array methods that could be helpful (indexOf, filter)
 */
 
-function getLanes() {}
+function getLanes(streets) {
+  let newArr = streets.filter((street) => street.includes("Lane"));
+  return newArr;
+}
```
Gayle-Thompson-Igwebike, line 12: Hi Natalie. I found your solution to mandatory 7 very helpful. I was struggling with the syntax, which is supposed to be very easy, but i just could not understand it. But yours made it very very easy for me to follow. Great job.

PR #187, opened 2022-12-16
Sana Asaf -London 9-Javascript-Core-1-Course week 4

SanaAsaf:
Exercises done and 2 mandatory done

### DIFF
--- a/2-mandatory/1-create-functions.js
+++ b/2-mandatory/1-create-functions.js
@@ -3,7 +3,8 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the first five elements of the passed array.
 */
-function first5() {
+function first5(array) {
+  return array.slice(0,5) ;
 }
 
 /*
@@ -11,7 +12,8 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(array) {
+return array.slice().sort();
 }
 
 /*
@@ -24,7 +26,8 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
+function tidyUpString(array) {
+  return array.map(value =>value.trim().replace('/', '').toLowerCase());
 }
 
 /*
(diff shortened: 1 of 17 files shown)


2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,17 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the first five elements of the passed array.
 */
-function first5() {
+function first5(array) {
+  return array.slice(0,5) ;
```
SallyMcGrath, line 7: 👍

2-mandatory/1-create-functions.js
```diff
@@ -33,7 +36,10 @@ Write a function that:
 - Returns a new array containing the same elements, but without the element at the passed index.
 */
 
-function remove() {
+function remove(array,index) {
+  let copyOfArray = Array.from(array);
```
SallyMcGrath, line 40:
```suggestion
  let copyOfArray = arr.slice();
```
Your approach will work, and this is a good solution. Array.from creates a shallow copy and this is what was asked for. 

https://developer.mozilla.org/en-US/docs/Glossary/Shallow_copy

People would typically use slice instead. 

There is a subtle difference between slice and Array.from that is completely non obvious and not actually relevant in this case, but is interesting to know about: slice makes a sparse array and Array.from does not. Later on, come back to this and read about it. For now, take your ⭐

2-mandatory/1-create-functions.js
```diff
@@ -44,7 +50,18 @@ Write a function that:
 - Numbers greater 100 must be replaced with 100.
 */
 
-function formatPercentage() {
+function formatPercentage(array) {
```
SallyMcGrath, line 53: Great job. This solution works well.

If you were to refactor your solution to use a map, how would you do that?

Start practising maps now as you will use them all the time in React.

2-mandatory/2-oxygen-levels.js
```diff
@@ -11,7 +11,12 @@
     Some string methods that might help you here are .replace() and .substring().
 */
 
-function findSafeOxygenLevel() {}
+function findSafeOxygenLevel(array) {
```
SallyMcGrath, line 14: This solution works well. I am finding it a bit hard to read. Do you have Prettier installed?

SallyMcGrath (commented):
Great job on the mandatory exercises that you submitted. Thanks so much for your hard work on this.

I think there are quite a few mandatory problems missing. Can you make sure to complete these and add them to this PR? Thanks!

2-mandatory/2-oxygen-levels.js
SanaAsaf: I cant read it  either.Appologies ! I didnt finish it and yes I have prettier installed .

PR #188, opened 2022-12-19
London_9_London_Lovelace_Meysam_Arshadi_JavaScript-Core-1-Coursework-Week4

Meysam-Arshadi:


1-exercises/J-string-substring/exercise2.js
```diff
@@ -14,11 +14,11 @@ let names = [
   "Arron Graham",
 ];
 
-names[0] = names[0].substring();
-names[1] = names[1].substring();
-names[2] = names[2].substring();
-names[3] = names[3].substring();
-names[4] = names[4].substring();
+names[0] = names[0].substring(0, 5);
```
SagharHosseinmardi, line 17: Hi Meysam just wanted to let you know in this line of code the indexes you used will print the name Tamzi instead of Tamzin the indexes for the print of full first name are (0, 6)

1-exercises/J-string-substring/exercise2.js
```diff
@@ -14,11 +14,11 @@ let names = [
...
+names[1] = names[1].substring(0, 6);
```
SagharHosseinmardi, line 18: In this line same if you want the full length of the first name the indexes are (0, 7).
nice try

PR #207, opened 2023-03-09
London Class10 | Seyyednavid HejaziJouybari | JavaScript-Core-1-Coursework | Week4

seyyednavid:


1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,8 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents = group.every(item => students.includes(item));
```
SeyoungJoo, line 8: Nice job 🙌

1-exercises/F-array-forEach/exercise.js
```diff
@@ -9,6 +9,26 @@
...
+    return 'Fizz';
+  } else if (item % 5 === 0){
+    return 'Buzz';
+  } else {
+    return item;
+  }
+}
+
+const result = (arr) => {
+  for (const item of arr) {
```
SeyoungJoo: This is also correct way to solve, but remember the constraint for this problem is using forEach method :)

2-mandatory/1-create-functions.js
```diff
@@ -24,16 +27,30 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
+function removeExtraThings(word) {
+  let newWord;
```
SeyoungJoo: In this case, you're not re-assigning any new value to newWord. so it's better just declair&assign at the same time 

`const newWord = word.trim().replace('/', '').toLowerCase();`

SeyoungJoo:
Super great job! you're now an expert of Array 🚀

1-exercises/C-array-every/exercise.js
seyyednavid: Thank you

1-exercises/F-array-forEach/exercise.js
seyyednavid: I corrected my code . Thank you for pointing it to me :pray:
const result = (arr) => {
  arr.forEach(item => {
    console.log(multiples(item));
  });

2-mandatory/1-create-functions.js
seyyednavid: function removeExtraThings(newword) {
  const newWord = newword.trim().replace('/', '').toLowerCase();
  return newWord;
}

seyyednavid:
> Super great job! you're now an expert of Array 🚀

Thank you Seyoung for your help and time :relaxed: :pray:

seyyednavid:
> Super great job! you're now an expert of Array 🚀

Thank you for your help and time :blush: :pray:

PR #208, opened 2023-03-11
complete instructions for oxygen levels exercise

shieldo:
The instructions for this exercise were not quite complete - some requirements were encoded into unit tests that were not mentioned in the original instructions.

I'm not sure whether they were intentionally incomplete, and that trainees are expected to look at failing tests to find hidden requirements for the exercise. This PR is making the assumption that actually we do want to include full requirements in the docblock before the function.


gregdyke (commented):
Having experienced this with trainees, what do you think about documenting this vs removing/changing some of the tests?

(the undefined one is kinda weird - I would probably have expected null? the validating of unexpected values just seems kinda annoying/gotcha-y?)

PR #209, opened 2023-03-11
LondonClass10_Bedrije_Omuri_Javascript1_Week4

Bedi06:


1-exercises/A-array-find/exercise.js
```diff
@@ -17,7 +17,9 @@ let names = [
   "Ahmed",
 ];
 
-let longNameThatStartsWithA = findLongNameThatStartsWithA(names);
+let char='A';
+
+let longNameThatStartsWithA = names.find((name)=>{return name.startsWith(char) && name.length>7})
```
seyyednavid: Hi @Bedi06. Nice Job :raised_hands: 
just a point. You can use callback function without return in the same line. I mean it returns result but it does not need return word in the same line.
names.find(name => name.startsWith('A') && name.length > 7);

Bedi06: Noted , Thank you Navid for your feedback will implement it .

PR #220, opened 2023-03-15
Glasgow Class 6 Melese Berehannu Javcript-Core-1-Coursework-Week4

Melesegithub:


2-mandatory/1-create-functions.js
```diff
@@ -24,16 +26,25 @@ Write a function that:
...
-
-function remove() {
+function remove(array, index) {
+  const withOutIndexElement = [];
+  for (let i = 0; i < array.length; i++) {
+    if (i !== index) {
+      withOutIndexElement.push(array[i]);
+    }
+  }
+  return withOutIndexElement;
```
RahmaB1, line 47: Brilliant  !! I think splice () can be used here as well.

2-mandatory/3-bush-berries.js
```diff
@@ -23,6 +23,14 @@
...
 function isBushSafe(berryArray) {
   //Write your code here
+  for (let berry of berryArray){
+    if (berry ==="pink"){
+     return "Bush is safe to eat from";
+    }else{
+     return "Toxic! Leave bush alone!";
+    }
+}
+
```
RahmaB1, line 33: Well Done !! I would say consider using every( ) or some ( ) as you'll need to make sure that the passed array contains only "pink". If it contains  other than "pink" then "Toxic..." should be returned.  Hope that helps.

RahmaB1:
Well done keep the good work 👌

PR #221, opened 2023-03-16
London class-10 -Junita Lama-JavaScript-1-week 4

Junitalama:


1-exercises/A-array-find/exercise.js
```diff
@@ -16,8 +16,11 @@ let names = [
   "Karim",
   "Ahmed",
 ];
+ //function longNameWithA(name) {
+ // return name.length > 7;
+//}
 
-let longNameThatStartsWithA = findLongNameThatStartsWithA(names);
```
jxz12, line 20: For this exercise you should define your own function `findLongNameThatStartsWithA` and implement your logic there.

1-exercises/A-array-find/exercise.js
```diff
@@ -16,8 +16,11 @@ let names = [
...
+let longNameThatStartsWithA = names.find(name => name.length > 7);
```
jxz12, line 23: You have the right idea here but just need to also add logic for starting with A

1-exercises/B-array-some/exercise.js
```diff
@@ -11,7 +11,11 @@ let pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 // If there is a null value in the array exit the program with the error code
 // https://nodejs.org/api/process.html#process_process_exit_code
 // process.exit(1);
-
+ for (let pair of pairsByIndex){
+  if (!pair){
```
jxz12, line 15: I think here it is better to use `pair === null` because that will check the type too

1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,19 @@
...
+return true;
+    } else{
+      return false;
+    }
+
+  }
+
+}*/
+
+let groupIsOnlyStudents = students.every(name => group.includes(students)? true : false);
```
jxz12, line 20: here you don't need the ternary operator `?` because the `.includes()` function already returns a bool

1-exercises/F-array-forEach/exercise.js
```diff
@@ -8,6 +8,25 @@
 */
 
 let arr = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
+arr.forEach(n => {
```
jxz12, line 11: your logic here is good but I think you should take more care to indent your code more neatly

1-exercises/H-array-methods-2/exercise.js
```diff
@@ -15,8 +15,8 @@ let everyone = [
   "Swathi",
 ];
 
-let firstFive; // complete this statement
-let lastFive; // complete this statement
+let firstFive = everyone.slice(0, 5); // complete this statement
+let lastFive = everyone.slice(2, 7); // complete this statement
```
jxz12, line 19: here you can use negative number indexes with `everyone.slice(-5)`

1-exercises/J-string-substring/exercise2.js
```diff
@@ -14,11 +14,11 @@ let names = [
...
-names[0] = names[0].substring();
-names[1] = names[1].substring();
-names[2] = names[2].substring();
-names[3] = names[3].substring();
-names[4] = names[4].substring();
+names[0] = names[0].substring(0,6);
+names[1] = names[1].substring(0,7);
+names[2] = names[2].substring(0,4);
+names[3] = names[3].substring(0,4);
+names[4] = names[4].substring(0,5);
```
jxz12, line 21: you could also do e.g. `names[4].split(' ')[1]` to solve this exercise, but your solution is good

1-exercises/J-string-substring/exercise3.js
```diff
@@ -6,10 +6,7 @@
   HINT: You will need to use .substring() twice
 */
 
-let statement = "I do not like programming";
```
jxz12, line 9: you could also use `.splice()` to nolve this if you turn it into an array first with `.split()` them `.join()`

2-mandatory/2-oxygen-levels.js
```diff
@@ -11,7 +11,17 @@
     Some string methods that might help you here are .replace() and .substring().
 */
 
-function findSafeOxygenLevel() {}
+function findSafeOxygenLevel(oxygenLevels) {
+let planets=oxygenLevels.filter(oxygenLevel =>oxygenLevel.includes("%"))
+let numberLevels= planets.map(number => parseFloat(number))
```
jxz12, line 16: I would like better indentation here

jxz12, line 16: here you could chain together the array methods without having to declare variables with `let`

2-mandatory/5-eligible-students.js
```diff
@@ -7,7 +7,10 @@
   - Returns an array containing only the names of the who have attended AT LEAST 8 classes
 */
 
-function getEligibleStudents() {}
+function getEligibleStudents(students, attendance) {
+  let studentName = students.filter(attendance=>attendance[1]>=8)
+  return studentName.map(names=>names[0]);
```
jxz12, line 12: here you could chain together filter and map with the extra `let`

2-mandatory/6-journey-planner.js
```diff
@@ -81,7 +84,12 @@ function getTransportModes() {}
 
   Hint: Use the corresponding array method to decide if an element is included in an array.
 */
-function isAccessibleByTransportMode() {}
+function isAccessibleByTransportMode(arr, str) {
+  if(arr.includes(str)){
```
jxz12, line 88: here you could just return the result of `.includes()` because it is already a boolean

2-mandatory/6-journey-planner.js
```diff
@@ -123,8 +133,11 @@ function getLocationName() {}
 */
 function journeyPlanner(locations, transportMode) {
   // Implement the function body
+let accessibleLocations= locations.filter( location=> location.includes(transportMode))
```
jxz12, line 136: you should use the `isAccessibleByTransportMode()` function above here, but your solution is also fine

2-mandatory/8-password-validator.js
```diff
@@ -23,11 +23,14 @@ PasswordValidationResult=  [false, false, false, false, true]
...
-function validatePasswords(passwords) {}
+     
+
+
+
 
 // Returns true if string contains at least one uppercase letter.
 function containsUppercaseLetter(string) {
-  return /[A-Z]/.test(string);
+  return string.test(string);
```
jxz12, line 33: this was already correct so no need to change it

2-mandatory/8-password-validator.js
```diff
@@ -23,11 +23,14 @@ PasswordValidationResult=  [false, false, false, false, true]
...
 
-function validatePasswords(passwords) {}
+     
+
+
+
 
 // Returns true if string contains at least one uppercase letter.
 function containsUppercaseLetter(string) {
-  return /[A-Z]/.test(string);
```
jxz12, line 30: we have not covered Regex much in class, but you should try to copy the example here. For example, a lower case test would be `/[a-z]/.test()` and a number would be `/[0-9]/.test()`

Junitalama:
@jxz12 Thank you  so much jonathan for reviewing. i will make changes as per your feedbacks.

PR #223, opened 2023-03-17
London-10-shadi-fakhri-JavaScript-Core-1-Coursework-Week4 

Shadi38:


2-mandatory/1-create-functions.js
```diff
@@ -24,16 +44,30 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
+ 
+function tidyUpString(arr) {
```
llamojha, line 48: Good job with this function. Just a minor comment, try to give more meaningful names to the variables and avoid using keywords like 'array' or 'string' as those can get confused with the type.

2-mandatory/8-password-validator.js
```diff
@@ -23,7 +23,31 @@ PasswordValidationResult=  [false, false, false, false, true]
...
 */
 
-function validatePasswords(passwords) {}
+function validatePasswords(passwords) {
+  let previousPasswords = [];
+  return passwords.map((password) => {
+    if (password.length < 5) {
+      return false;
+    }
+    if (!containsUppercaseLetter(password)) {
```
llamojha, line 32: You have multiple if statements that will do the same (return false;). How would you optimize this code?

PR #226, opened 2023-03-20
Glasgow Class 6 - Man Sang Sin - JavaScript Core 1 - Week 4

ManSangSin:


2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,19 @@ Write a function that:
...
 /*
 Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(inputArray) {
+  let sortedArr = inputArray.sort();
+  return sortedArr;
 }
```
RahmaB1, line 19: Well done !! 
Please be aware that sort() method sorts the elements of an array and returns the reference to the same array. 
Which means it will sort inputArray. To avoid this, you can try to copy the inputArray as value then sort the new array. 
sortedArr = [...inputArray]; //it will copy the value of inputArray and not a reference to it. 
sortedArr.sort( ); 

 [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort](url)
 
 Hope that helps. 😊
 Keep up the good work 👌👌

PR #230, opened 2023-03-25
LONDON_10 || SAIM KORKMAZ || JS1-4

nsaimk:
Mentor: @berkeli 
Mates: @mabalal , @Mr-DEM1R

### DIFF
--- a/2-mandatory/1-create-functions.js
+++ b/2-mandatory/1-create-functions.js
@@ -4,6 +4,7 @@ Write a function that:
 - Returns a new array containing the first five elements of the passed array.
 */
 function first5(arr) {
+  return arr.filter((element, index) => index < 5);
 }
 
 /*
@@ -11,7 +12,8 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(arr) {
+  return arr.sort();
 }
 
 /*
@@ -24,7 +26,8 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
+function tidyUpString(arr) {
+  return arr.map((str) => str.trim().replace(/\//g, "").toLowerCase());
 }
 
 /*
@@ -33,7 +36,9 @@ Write a function that:
 - Returns a new array containing the same elements, but without the element at the passed index.
 */
 
-function remove() {
+function remove(arr, index) {
+  const newArr = arr.filter((v, i) => index == i);
+  return newArr;
 }
 
 /*
@@ -44,7 +49,14 @@ Write a function that:
 - Numbers greater 100 must be replaced with 100.
 */
 
-function formatPercentage() {
+function formatPercentage(arr) {
+  const formattedPercentages = numbers.map((number) => {
+    if (number > 100) {
+      number = 100;
+    }
+    return (Math.round(number * 100) / 100).toFixed(2) + "%";
+  });
+  return formattedPercentages;
 }
 
 /* ======= TESTS - DO NOT MODIFY ===== */
--- a/2-mandatory/1-create-functions.js
+++ b/2-mandatory/1-create-functions.js
@@ -3,7 +3,7 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the first five elements of the passed array.
 */
-function first5() {
+function first5(arr) {
 }
 
 /*
--- a/2-mandatory/3-bush-berries.js
+++ b/2-mandatory/3-bush-berries.js
@@ -23,11 +23,19 @@
 
 function isBushSafe(berryArray) {
   //Write your code here
-}
+   if (berryArray.every(v => v == "pink")) {
+    return "Bush is safe to eat from";
+  } else {
+    return "Toxic! Leave bush alone!";
+  } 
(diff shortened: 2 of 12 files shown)


2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,17 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the first five elements of the passed array.
 */
-function first5() {
+function first5(arr) {
+  return arr.filter((element, index) => index < 5);
```
berkeli, line 7: This is not a good approach, filtering is an expensive operation that iterates over each element in the array. If array has `n` elements, it will result in `n` operations. 

You can get the first 5 elements with 1 operation - `arr.slice(0, 5);`

2-mandatory/1-create-functions.js
```diff
@@ -33,7 +36,9 @@ Write a function that:
 - Returns a new array containing the same elements, but without the element at the passed index.
 */
 
-function remove() {
+function remove(arr, index) {
+  const newArr = arr.filter((v, i) => index == i);
```
berkeli, line 40: Similar comment as before, we should not use filter if we know which element we don't need:
```
const removedElement = arr.splice(index,1)
return arr // this will not contain removedElement
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

This modifies the source array, so if you want to avoid that - you can create a copy via` const arr1 = [...arr];`

2-mandatory/1-create-functions.js
```diff
@@ -44,7 +49,14 @@ Write a function that:
 - Numbers greater 100 must be replaced with 100.
 */
 
-function formatPercentage() {
+function formatPercentage(arr) {
+  const formattedPercentages = numbers.map((number) => {
```
berkeli, line 53:
```suggestion
  const formattedPercentages = arr.map((number) => {
```

numbers is not defined here and also your tests are not passing for this function. 

I'd suggest using `Math.round(num * 100) / 100` instead of `toFixed()`

2-mandatory/1-create-functions.js
```diff
@@ -44,7 +49,14 @@ Write a function that:
...
+    if (number > 100) {
+      number = 100;
+    }
+    return (Math.round(number * 100) / 100).toFixed(2) + "%";
```
berkeli, line 57:
```suggestion
    return (Math.round(number * 100) / 100)+ "%";
```

toFixed is not needed here, it will already round to 2 decimal places.

2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,17 @@ Write a function that:
...
 
 /*
 Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(arr) {
+  return arr.sort();
 }
```
berkeli, line 17: Some tests for this function are failing because it modifies the passed in array. I'd suggest making a copy of the array and sorting that, e.g `[...arr].sort();`

2-mandatory/3-bush-berries.js
```diff
@@ -23,11 +23,19 @@
 
 function isBushSafe(berryArray) {
   //Write your code here
-}
+   if (berryArray.every(v => v == "pink")) {
+    return "Bush is safe to eat from";
+  } else {
+    return "Toxic! Leave bush alone!";
+  }
```
berkeli, line 30:
```suggestion
   if (berryArray.every(v => v == "pink")) {
    return "Bush is safe to eat from";
  }
  return "Toxic! Leave bush alone!"; 
```
We can skip the else statement for a cleaner code

2-mandatory/4-space-colonies.js
```diff
@@ -15,7 +15,11 @@
   
 */
 
-function getSettlers() {}
+function getSettlers(arr) {
+  return (result = arr.filter(
+    (v) => v.startsWith("A") && v.includes("family")
+  ));
```
berkeli, line 21: Can you explain why you need result here?

2-mandatory/5-eligible-students.js
```diff
@@ -7,7 +7,17 @@
...
+  let result = [];
+
+  for (let i = 0; i < arr.length; i++) {
+    if (arr[i][1] > 7) {
+      result.push(arr[i][0])
+    }
+  }
+
+  return result
+}
```
berkeli, line 20: This works fine, but since this week was focused on array methods - it would be nice if you can rewrite with `.filter()` and `.map()`

2-mandatory/6-journey-planner.js
```diff
@@ -123,6 +129,13 @@ function getLocationName() {}
 */
 function journeyPlanner(locations, transportMode) {
   // Implement the function body
+  let result = [];
+  for (let i = 0; i < locations.length; i++) {
+    if (locations[i].includes(transportMode)){
+      result.push(locations[i][0])
+    }
+  }
+  return result
```
berkeli, line 138: I think the idea here was to use the functions you wrote earlier (`getLocationName` and `isAccessibleByTransportMode`)

also, you can use` .filter() `and `.map()`

berkeli (commented):
Nice work, but I see there are multiple exercises missing in the 1-exercises folder.

PR #233, opened 2023-04-01
Clean up unnecessary exercises

### LINKED ISSUE #502: JS1 Week 4 work review
Which module(s) and week(s) does this change affect?
Module(s): JS1
Week(s): 4

What is the work that needs to be done?
https://github.com/CodeYourFuture/syllabus/issues/494

Why is this work important to do?

Additional context

Who might need to know about this change?

40thieves:
Closes https://github.com/CodeYourFuture/syllabus/issues/502.

## Summary of changes

- Removes the in-class `exercises` code & content (which doesn't match the syllabus any more)
- Improves test instructions
- Switches an extra exercise from a markdown file to a JS file (I don't see much value in getting trainees to create the file)
- Removes an exercise that isn't related to the lesson any more
- Adds some more mandatory exercises (see below)

## More mandatory exercises

I noticed that the mandatory exercises don't cover all of the content from the lesson (e.g. some array methods weren't used by the tests), so I've moved some of the in-class exercises into the mandatory folder. I did some minimal work to convert them to actual tests instead of just expected output comments.

Given that the purpose of this work was to _reduce_ the workload, I'm not 100% convinced this is the best choice - so I've included these as the last two commits (4df079d63efca05229a74fa57c5b70d9afb72b3a and 4ece0577ba47bac81437b2df71d4255362bdca9e). I'm happy to drop these commits if we feel that they aren't necessary.


mandatory/1-create-functions.js
```diff
@@ -3,16 +3,14 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the first five elements of the passed array.
 */
-function first5() {
-}
+function first5() {}
```
Dedekind561, line 6: Please can we get rid of this function too whilst we're here?

mandatory/10-capitalize.js
```diff
@@ -0,0 +1,28 @@
...
+/* 
+===================================================
+======= TESTS - DO NOT MODIFY BELOW THIS LINE =====
+There are some Tests in this file that will help you work out if your code is working.
+To run the tests for just this one file, type `npm test -- --testPathPattern 10-capitalize` into your terminal
+(Reminder: You must have run `npm install` one time before this will work!)
+===================================================
+*/
+
+test("Capitalize works - case 1", () => {
```
Dedekind561:
```suggestion
test("capitalize works - case 1", () => {
```

mandatory/12-fizz-buzz.js
```diff
@@ -0,0 +1,44 @@
...
+function fizzBuzz(numbers) {}
+
+/* 
+===================================================
+======= TESTS - DO NOT MODIFY BELOW THIS LINE =====
+There are some Tests in this file that will help you work out if your code is working.
+To run the tests for just this one file, type `npm test -- --testPathPattern 12-fizz-buzz` into your terminal
+(Reminder: You must have run `npm install` one time before this will work!)
+===================================================
+*/
```
Dedekind561: Do we need additional test cases for smaller values here too?

mandatory/10-capitalize.js
```diff
@@ -0,0 +1,28 @@
...
+  expect(capitalize("hello")).toEqual("Hello");
+});
+
+test("Capitalize works - case 2", () => {
```
Dedekind561:
```suggestion
test("capitalize works - case 2", () => {
```

mandatory/10-capitalize.js
```diff
@@ -0,0 +1,28 @@
...
+  expect(capitalize("daniel")).toEqual("Daniel");
+});
+
+test("Capitalize works - case 2", () => {
```
Dedekind561:
```suggestion
test("capitalize works - case 3", () => {
```

mandatory/9-group-people.js
```diff
@@ -0,0 +1,33 @@
...
+/* 
+===================================================
+======= TESTS - DO NOT MODIFY BELOW THIS LINE =====
+There are some Tests in this file that will help you work out if your code is working.
+To run the tests for just this one file, type `npm test -- --testPathPattern 2-oxygen-levels` into your terminal
+(Reminder: You must have run `npm install` one time before this will work!)
+===================================================
+*/
+
+TestScheduler("volunteersAndTrainees works", () => {
```
Dedekind561: I'm assuming this is an auto-complete error?

mandatory/4-space-colonies.js
```diff
@@ -17,7 +17,14 @@
 
 function getSettlers() {}
 
-/* ======= TESTS - DO NOT MODIFY ===== */
+/*
```
Dedekind561: I don't think this exercise is appropriate and I think it should be binned - what do you think?

Dedekind561 (changes requested):

mandatory/1-create-functions.js
alismith-dev: I'm not sure I follow - are you asking to remove this (sub-)exercise? Or to remove the stubbed function?

If you're asking for the latter, I think we do need it as we call this specific named function in the tests, so getting trainees to write it themselves risks misnaming, which could lead to confusing errors. I guess I would be fine with going down that route, but that feels like a policy decision that should be taken across the board.

mandatory/4-space-colonies.js
alismith-dev: Yeah, I can see that. I'll remove it

mandatory/9-group-people.js
alismith-dev: Yep, my vscode has this annoying habit of auto-selecting the first suggestion and I'm not sure why. Will fix.

mandatory/12-fizz-buzz.js
alismith-dev: Sorry, I'm not sure I follow what you're asking here?

Is your question that we're not testing for example `fizzBuzz([1, 2, 3])`? If so, then I can see it both ways: the existing test has assertions for all of those values, but we'd lack good test descriptions for those. Personally I would probably add more individual test cases with explicit test descriptions but that didn't seem to be the style in other test files in the coursework.

PR #236, opened 2023-04-05
London10-Afsha-Hossain-JS1-Week4

Afsha10:


1-exercises/A-array-find/exercise.js
```diff
@@ -17,9 +20,46 @@ let names = [
   "Ahmed",
 ];
 
+function isLongName(singleName) {
+  return singleName.length > 7;
+}
+
+let longName = names.find(isLongName);
```
kitko112, line 27: Good use of `find` function. FYI it is also possible to write line 23-27 into one statement. The benefit is that you can save the effort on thinking a function name `isLongName`.  It is a personal coding style, not a rule to follow though.
```
let longName = names.find((name) => singleName.length > 7);
```

1-exercises/A-array-find/exercise.js
```diff
@@ -17,9 +20,46 @@ let names = [
...
+
+function isStartingWithA(singleName) {
+  return singleName[0] === "A";
+}
+
+let nameStartingWithA = names.find(isStartingWithA);
+
+function findLongNameThatStartsWithA(names) {
```
kitko112, line 35: The `names` parameter is being used inside the function, do you think it can be removed?

To write the code simpler, on line 40 we can do the comparison directly as there is just one line of code.
```
let longNameThatStartsWithA = nameStartingWithA && longName;
```

1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,11 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents = group.every(isOnlyStudent); // complete this statement
+
+function isOnlyStudent(list) {
+  return list.includes(students);
```
kitko112, line 11: Good use of every. What is the value of `list`? is it the element of group like "Austine", "Dany"... ? 
on line 11 if we want to check if the name from the group is included in students array, it should be like `students.includes` right?

1-exercises/G-array-methods/exercise2.js
```diff
@@ -7,7 +7,9 @@
 let mentors = ["Daniel", "Irina", "Rares"];
 let students = ["Rukmini", "Abdul", "Austine", "Swathi"];
 
-let everyone; // complete this statement
+
+let everyone = mentors.concat(students); // complete this statement
```
kitko112, line 11: Good use of `concat`, FYI to combine two arrays it can also be done as below
```
let everyone = [...mentors, ...students];
```

1-exercises/I-string-replace/exercise.js
```diff
@@ -13,7 +13,8 @@
 let story =
   "I like dogs. One day I went to the park and I saw 10 dogs. It was a great day.";
 
-let result = story.replace("", "");
+let result =
+  story.replace("dogs", "cats").replace("day", "night").replace("10", "100000").replace("dogs", "cats").replace("great", "brilliant").replace("day", "night");
```
kitko112, line 17: Good use of replace, FYI there is a way to replace all the occurrence with one `replace` function, in this case we want to replace `dogs` into ` cats`, it can be done by regular expression with the global flag `g`. If you are interested you can look it up. 
```
story.replace(/dogs/g, 'cats')
```

2-mandatory/1-create-functions.js
```diff
@@ -24,7 +26,11 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
+function tidyUpString(arrayOfStrings) {
+  for (const singleString of arrayOfStrings) {
+    console.log(singleString.trim());
+  }
+  return singleString.trim().replace("/","").toLowerCase();
```
kitko112, line 33: is `singleString` accessible here still?

2-mandatory/5-eligible-students.js
```diff
@@ -7,7 +7,14 @@
   - Returns an array containing only the names of the who have attended AT LEAST 8 classes
 */
 
-function getEligibleStudents() {}
+function findName(studentArray) {
+  let filterArray = studentArray.filter((element) => {(element[1] >= 8) {
```
kitko112, line 11: It seems the `if` is missing

2-mandatory/7-lane-names.js
```diff
@@ -6,7 +6,15 @@
   HINT: string and array methods that could be helpful (indexOf, filter)
 */
 
-function getLanes() {}
+let laneNames = [];
+
+
+function getLanes(roadNames) {
+  if (roadNames.includes("Lane")) {
```
kitko112, line 13: As the hint said, indexOf would be useful to help check the partial match in the string.

PR #237, opened 2023-04-09
Glasgow-6_JS1-Week-4_Ehdaa-Sakawi

Ehdaa-Munier:


2-mandatory/1-create-functions.js
```diff
@@ -24,16 +33,26 @@ Write a function that:
...
 - Returns a new array containing the same elements, but without the element at the passed index.
 */
+//
+
+
 
-function remove() {
+function remove(array, index) {
+ 
+return array.slice(0, index).concat(array.slice(index + 1));
```
annacollins85, line 55: Nice way of doing it

annacollins85, line 55: Nice way of doing it 👏

2-mandatory/1-create-functions.js
```diff
@@ -44,7 +63,17 @@ Write a function that:
 - Numbers greater 100 must be replaced with 100.
 */
 
-function formatPercentage() {
+function formatPercentage(numbers) {
+  numbersWithPercentages = numbers.map(addPercentage);
+   return numbersWithPercentages;
```
annacollins85, line 68: Here we could simply return the map directly without saving it in a variable and then returning that.
`numbers.map(addPercentage);`

Good use of map though 👏

2-mandatory/2-oxygen-levels.js
```diff
@@ -10,8 +10,17 @@
 
     Some string methods that might help you here are .replace() and .substring().
 */
+let oxegenLevel= [];
+function findSafeOxygenLevel(result) {
+  if(value > 19.5 && value < 23.5)
+  return result === "%", " . ";
+}
```
annacollins85, line 18: Here the function should take an array of percentages as an input (see tests for examples below). So first we need to turn the percentage which is currently a strings into a number. We need to find the first number in the array that is between 19.5 and 23.5. `parseFloat` will turn a percentage into a number, e.g. `parseFloat("25%") === 25`

```js
function findSafeOxygenLevel(planetPercentages) {
  return planetPercentages.find(function(percentage) {
    const number = percentage.parseFloat(percentage)
    return (number >= 19.5 && number <= 23.5) // This will return true or false, if it returns true, the `find` array function will return it as the element it finds first
  })
}
```

2-mandatory/3-bush-berries.js
```diff
@@ -23,8 +23,15 @@
 
 function isBushSafe(berryArray) {
   //Write your code here
+  
+  if (berryArray.every((berry) => berry === "pink")) {
+    return  "Bush is safe to eat from";
+  } else {
+    return "Toxic! Leave bush alone!";
+  }
```
annacollins85, line 31: Great!

2-mandatory/4-space-colonies.js
```diff
@@ -15,8 +15,17 @@
...
+function getSettlers(voyagers) {
 
+ let array=[];
+ let array1= [ array+"family" , "family"];
+ for(let familyName of voyagers){
+  if(familyName.split(' ').slice(-1).join(' ') && familyName[0] === "A"){
+  array.push(familyName)}
+}
+
+return array1;
```
annacollins85, line 27: You could use filter here. (Note: I have not run this code so haven't checked syntax)
```js
function getSettlers(voyagers) {
  return  voyagers.filter(function(voyager) {
     // separate the name into an array of strings
    const names = voyager.split(' ');
    // check the name includes the string "family", the first word starts with "a"
    if (names.includes("family") && names[0].startsWith("a")) {
       return true
    } 
    return false
  })
}
```

2-mandatory/5-eligible-students.js
```diff
@@ -7,8 +7,12 @@
   - Returns an array containing only the names of the who have attended AT LEAST 8 classes
 */
 
-function getEligibleStudents() {}
+function getEligibleStudents(attended) {
 
+ const eligibleAttendance = attended.filter(attendance => attendance[1] >= 8);
+ return eligibleAttendance.map(names=> names[0])
```
annacollins85, line 13: Great 👏

2-mandatory/7-lane-names.js
```diff
@@ -6,7 +6,9 @@
   HINT: string and array methods that could be helpful (indexOf, filter)
 */
 
-function getLanes() {}
+function getLanes(streetNames) {
+    return streetNames.filter((street) => street.includes("Lane"));
+}
```
annacollins85, line 11: 👏

2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,24 @@ Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the first five elements of the passed array.
 */
-function first5() {
+function first5(Arr) {
+//let numbers = [ 1, 2, 3, 4, 5, 6, 7]
+let copyOfOriginal = Arr.slice(0, 5); {
+   return copyOfOriginal;
```
annacollins85, line 9: 👏

2-mandatory/1-create-functions.js
```diff
@@ -3,15 +3,24 @@ Write a function that:
...
 /*
 Write a function that:
 - Accepts an array as a parameter.
 - Returns a new array containing the same elements, except sorted.
 */
-function sortArray() {
+function sortArray(letters) {
+  let copyOfOriginal = [...letters];
+  copyOfOriginal.sort();
+  return copyOfOriginal;
```
annacollins85, line 23: This is good. We could also have just directly returned the sorted version like this
```suggestion
function sortArray(letters) {
  let copyOfOriginal = [...letters];
  return copyOfOriginal.sort();
  }
```

2-mandatory/1-create-functions.js
```diff
@@ -24,16 +33,26 @@ Write a function that:
 - Removes any forward slashes (/) in the strings.
 - Makes the strings all lowercase.
 */
-function tidyUpString() {
+function tidyUpString(array) {
+ 
+
+  return array.map((string) => string.trim().toLowerCase().replace("/", ""));
```
annacollins85, line 39: Good use of method chaining! 👏

annacollins85 (commented):
Well done! I have added some comments and some answers to some of the questions. I see you had a little trouble with `6-journey-planner.js`. Maybe we can have a look at this in a buddy session. Or I advise you to ask some of your classmates about how they implemented it when you're in class on Saturday.

PR #239, opened 2023-04-13
London10 - Rihanna P - JJavaScript-Core-1-Coursework-Week4

RihannaP:


1-exercises/B-array-some/exercise.js
```diff
@@ -15,10 +15,12 @@ let pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 let students = ["Islam", "Lesley", "Harun", "Rukmini"];
 let mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
 
+function isNull(input){
+  return input === null};
```
GergelyKI: The solutions look good regarding the logic. 

If I can add anything, it'd be to pay a bit more attention to code styling, like in the line above the `};` could have gone in a new line. At some other places, it's also a bit inconsistent where you use spaces or semicolons.

However, like I mentioned it's only styling, it's more important when you work with others :)

PR #240, opened 2023-04-19
London10-StellaDelMar_RodriguezFernandez-JavaScript-Core-1-Coursework-Week4

Stelladelmar:
will do mandatory

### DIFF
--- a/1-exercises/B-array-some/exercise.js
+++ b/1-exercises/B-array-some/exercise.js
@@ -8,9 +8,18 @@
 
 let pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 
+//***************************************************
+
+function isNullvalue(!number) {
+  return !number;
+}
+
+//***************************************************
+
 // If there is a null value in the array exit the program with the error code
 // https://nodejs.org/api/process.html#process_process_exit_code
 // process.exit(1);
+
 
 let students = ["Islam", "Lesley", "Harun", "Rukmini"];
 let mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
--- a/1-exercises/C-array-every/exercise.js
+++ b/1-exercises/C-array-every/exercise.js
@@ -5,7 +5,11 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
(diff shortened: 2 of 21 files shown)


1-exercises/B-array-some/exercise.js
```diff
@@ -8,10 +8,19 @@
 
 let pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 
+//***************************************************
+
+function isNullvalue(!number) {
```
ButcherDing, line 13: usually we don't use ! in our arguments

1-exercises/B-array-some/exercise.js
```diff
@@ -8,10 +8,19 @@
...
+  return !number;
+}
+
+//***************************************************
+
```
ButcherDing, line 18: Usually we don't use ! in our arguments. Also the error code still needs to be returned (in the comments below).

1-exercises/C-array-every/exercise.js
```diff
@@ -5,7 +5,11 @@
 let students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 let group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-let groupIsOnlyStudents; // complete this statement
+let groupIsOnlyStudents = group.every(any);{
+return students.includes(any);
+}
+console.log(groupIsOnlyStudents);
+
```
ButcherDing, line 12: 'any' isn't defined anywhere, and maybe rethink what is going to go inside .every().
.every() takes an array and gives us back 'true' if every element passes a certain test - this test can be a function - It looks like this is what you were trying to do.

1-exercises/D-array-filter/exercise.js
```diff
@@ -8,7 +8,10 @@
 
 let pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-let pairsByIndex; // Complete this statement
+let pairsByIndex = pairsByIndexRaw.filter( item => item.length === 2);
+
+
+// Complete this statement
```
ButcherDing, line 15: Nice!

1-exercises/C-array-every/exercise.js
Stelladelmar: Thank you 🙏 I see

